PR #29, opened 2022-01-01
Update dependencies

Trouv:
Getting texture array support


Trouv:
There are a couple issues with the current texture atlas support that I'd like to be resolved before merging this.
- [ ] Tile flips don't seem to work anymore
- [ ] Some strange artifacts on the edges of transparent tiles

Issue #39, opened 2022-01-09
Regressions with tile spacing


Trouv:
Since updating to bevy 0.6/bevy_ecs_tilemap 0.5, loading ldtk files with AutoTile and Tile layers using tileset spacing can lead to crashes.

This may be an upstream issue, and if it is, it's probably not a good idea to hold up a 0.1 release for it. If it's the only choice for getting 0.1 out the door, I'm okay not supporting tileset spacing for AutoTile and Tile layers, so long as it doesn't lead to crashes (maybe a warning?)

PR #54, opened 2022-02-12
Add Atlas Pass-Through Feature to Support Wasm

colathro:


Trouv:
Thanks so much! Tested, merging.
Note: this has implications for #39 as well

PR #80, opened 2022-04-24
Tile data and refactors

### LINKED ISSUE #52: Access tile data
Hi, 

Ldtk makes it possible to link data to tiles of a tile-set (either arbitrary text or enum-based)

It would be nice if there was a `TileData`(or equivalent) component inserted on tiles that have data defined so that we could query them and have logic based on the data we defined in ldtk.

Trouv:
Closes #52

## New components
Adds `TileMetadata` and `TileEnumTags` components. These get added to any tile whose corresponding tile in the LDtk tileset definitions has metadata or enum tags.

This functionality applies to IntGrid layers with auto-tile rules, Tile layers, and AutoTile layers. On all these layers, the resulting entities are bevy_ecs_tilemap Tiles w/ `GridCoords`, and will also receive correct Transforms, just like normal intgrid tiles do. The Transforms are NOT added to all tiles, just the ones that either have intgrid values or metadata.

If it's a Tile/AutoTile layer, and that layer's tileset definition does not have any metadata, that layer will be built with the performant `LayerBuilder::new_batch`. Otherwise, the slower `LayerBuilder` functions are used to build layers. Users can "opt-out" of this small performance hit by just not adding metadata to their tilesets in LDtk.

## Refactors
Went a little wild with the refactors here...

### conversion functions
Cleaned up the coordinate-conversion utility functions quite a bit. These changes embrace `GridCoords` a bit more. Pretty much any conversions around `TilePos` have now been replaced by `GridCoords`. This isn't too painful of a breaking change, since `GridCoords` is `From<TilePos>` and `TilePos` is now `From<GridCoords>`. It also adds some conversions that were missing, like bevy-space translation to `GridCoords`.

The result is the same number of conversion functions (one of them being `From<GridCoords> for TilePos`), but now the selection is a bit more organized and more useful.

### level spawning
Added a new `level` module just for the `spawn_level` function, and split up said function quite a bit more than it was. This reduced a good amount of repeated code and cleaned up other modules that were being a little overwhelmed by level-spawning (`systems` and `utils` in particular).


Trouv:
After submitting, realized this could probably use an example.
Might also want to look into doing quickly iterating over a layer to see if any of its tiles are even in the metadata_map/enum_tags_map before deciding not to use `new_batch`, but it might not be worth it.

But not today

Trouv:
Decided to split the example writing into its own issue. Don't think it should hold up the 0.3 release. #82

PR #84, opened 2022-04-26
Add Reflect derives for components

johanhelsing:
This makes the types easier to work with when serializing save games or implementing network rollback.

With this patch, it works with https://github.com/johanhelsing/bevy_snap


Trouv:
Looks good to me. Thanks for the contribution

Issue #44, opened 2022-01-16
Ground detection for platformer example


Trouv:
Currently, you can jump at any time in the platformer example. Make it so you can only jump when you're on the ground.

PR #86, opened 2022-04-28
Implement Ground detection in platformer example

ChristopherBiscardi:
Restrict jumping to only happening when ground detection through shape casting with a filter says the player is on the ground.

This works for normal platforms. It ignores climbables by filtering them out of the shape cast.

You can still multi-jump up walls with this solution and I'm not sure how to fix that. The multi-jump up walls is caused by the wider shape cast than the character, which results in a nice coyote jump when leaving ledges. Shortening the shape does remove the "multi-wall-jump" effect (and also the coyote effect), so I assume it's because the shape cast is actually *in* the wall.

related to #44

### DIFF
--- a/examples/platformer/systems.rs
+++ b/examples/platformer/systems.rs
@@ -430,43 +430,60 @@ pub fn update_level_selection(
 
 pub fn spawn_ground_sensor(
     mut commands: Commands,
-    detect_ground_for: Query<(Entity, &CollisionShape), Added<GroundDetection>>,
+    detect_ground_for: Query<(Entity, &CollisionShape, &Transform), Added<GroundDetection>>,
 ) {
-    for (entity, shape) in detect_ground_for.iter() {
-        commands.entity(entity).with_children(|builder| {
-            builder
-                .spawn()
-                .insert(RigidBody::Sensor)
-                .insert(shape.clone())
-                .insert(Transform::from_xyz(0.0, -10.0, 5.0))
-                .insert(GroundSensor(entity));
-        });
+    for (entity, shape, transform) in detect_ground_for.iter() {
+        if let CollisionShape::Cuboid { half_extends, .. } = shape {
+            let detector_shape = CollisionShape::Cuboid {
+                half_extends: Vec3::new(half_extends.x / 2., 2., 0.),
+                border_radius: None,
+            };
+
+            commands.entity(entity).with_children(|builder| {
+                builder
+                    .spawn()
+                    .insert(RigidBody::Sensor)
+                    .insert(detector_shape)
+                    .insert(Transform::from_xyz(0.0, -half_extends.y - 4., 0.))
+                    .insert(GlobalTransform::default())
+                    .insert(GroundSensor {
+                        ground_detection_entity: entity,
+                        intersecting_ground_entities: HashSet::new(),
+                    });
+            });
+        }
     }
 }
 
 pub fn ground_detection(
-    mut ground_detectors: Query<(Entity, &mut GroundDetection)>,
-    ground_sensors: Query<(Entity, &GroundSensor)>,
+    mut ground_detectors: Query<&mut GroundDetection>,
+    mut ground_sensors: Query<(Entity, &mut GroundSensor)>,
     mut collisions: EventReader<CollisionEvent>,
 ) {
-    for sensor in ground_sensors.iter() {
+    for (entity, mut ground_sensor) in ground_sensors.iter_mut() {
         for collision in collisions.iter() {
             match collision {
-                CollisionEvent::Started(a, _) => {
-                    if a.rigid_body_entity() == sensor.0 {
-                        if let Ok((_, mut detector)) = ground_detectors.get_mut(sensor.1 .0) {
-                            detector.on_ground = true;
-                        }
+                CollisionEvent::Started(a, b) => {
+                    if a.rigid_body_entity() == entity {
+                        ground_sensor
+                            .intersecting_ground_entities
+                            .insert(b.rigid_body_entity());
                     }
                 }
-                CollisionEvent::Stopped(a, _) => {
-                    if a.rigid_body_entity() == sensor.0 {
-                        if let Ok((_, mut detector)) = ground_detectors.get_mut(sensor.1 .0) {
-                            detector.on_ground = false;
-                        }
+                CollisionEvent::Stopped(a, b) => {
+                    if a.rigid_body_entity() == entity {
+                        ground_sensor
+                            .intersecting_ground_entities
+                            .remove(&b.rigid_body_entity());
                     }
                 }
             }
+        }
+
+        if let Ok(mut ground_detection) =
+            ground_detectors.get_mut(ground_sensor.ground_detection_entity)
+        {
+            ground_detection.on_ground = ground_sensor.intersecting_ground_entities.len() > 0;
         }
     }
 }
--- a/examples/platformer/systems.rs
+++ b/examples/platformer/systems.rs
@@ -49,11 +49,10 @@ pub fn dbg_player_items(
 
 pub fn movement(
     input: Res<Input<KeyCode>>,
-    mut query: Query<(&mut Velocity, &mut Climber, &Transform), With<Player>>,
-    physics_world: PhysicsWorld,
-    climbables: Query<&Climbable>,
+    mut query: Query<(&mut Velocity, &mut Climber), With<Player>>,
+    ground_detection: Res<GroundDetection>,
 ) {
-    for (mut velocity, mut climber, transform) in query.iter_mut() {
+    for (mut velocity, mut climber) in query.iter_mut() {
         let right = if input.pressed(KeyCode::D) { 1. } else { 0. };
         let left = if input.pressed(KeyCode::A) { 1. } else { 0. };
 
@@ -72,33 +71,9 @@ pub fn movement(
             velocity.linear.y = (up - down) * 200.;
         }
 
-        if input.just_pressed(KeyCode::Space) {
-            let shape_size = Vec2::new(30., 1.);
-
-            let shape = CollisionShape::Cuboid {
-                half_extends: shape_size.extend(0.) / 2.,
-                border_radius: None,
-            };
-
-            let mut floor = transform.clone();
-            floor.translation.y = floor.translation.y - 20.0;
-
-            let result = physics_world.shape_cast_with_filter(
-                &shape,
-                floor.translation,
-                Quat::IDENTITY,
-                transform.translation,
-                CollisionLayers::default(),
-                |entity| climbables.get(entity).is_err(),
-            );
-
-            if let Some(collision) = result {
-                if let ShapeCastCollisionType::Collided(_) = collision.collision_type {
-                } else {
-                    velocity.linear.y = 450.;
-                    climber.climbing = false;
-                }
-            }
+        if input.just_pressed(KeyCode::Space) && ground_detection.on_ground {
+            velocity.linear.y = 450.;
+            climber.climbing = false;
         }
     }
 }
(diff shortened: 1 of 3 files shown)


Trouv:
Thanks for the contribution!

This works pretty well, though it does have the wall climbing issue you mentioned. I think I'd rather have no wall climbing than have coyote time. This preference is mostly because I think in a full-fledged platformer coyote time would be time-based rather than space-based, so the coyote time implementation here isn't necessarily setting a good example.

examples/platformer/systems.rs
```diff
@@ -68,8 +73,32 @@ pub fn movement(
...
+                &shape,
+                floor.translation,
+                Quat::IDENTITY,
+                transform.translation,
+                CollisionLayers::default(),
+                |entity| climbables.get(entity).is_err(),
+            );
+
+            if let Some(collision) = result {
+                if let ShapeCastCollisionType::Collided(_) = collision.collision_type {
```
Trouv, line 96: I think it would be better to perform the ground detection in a separate system. Maybe give the player a new `GroundDetection { on_ground: bool }` component, and set it to true in the other system when they're on the ground.

I think this gives a nice separation of concerns, allows for other entities to also detect the ground if they need to, and avoids accessing the `PhysicsWorld`, at least here.

examples/platformer/systems.rs
```diff
@@ -68,8 +73,32 @@ pub fn movement(
...
+
+            let shape = CollisionShape::Cuboid {
+                half_extends: shape_size.extend(0.) / 2.,
+                border_radius: None,
+            };
+
+            let mut floor = transform.clone();
+            floor.translation.y = floor.translation.y - 20.0;
+
+            let result = physics_world.shape_cast_with_filter(
```
Trouv, line 86: I think this method for ground detection works, but I did think of another way.

I believe that the player is constantly being pushed into the ground when they're not on a ladder due to the gravity, which is constantly firing `CollisionEvent`s. So, we could use an `EventReader<CollisionEvent>` to read these, similar to the ladder logic. I think that the `CollisionData::normals` may help us determine that what we're colliding with *is* the ground, as a collision normal that's just `Vec3::UP` would be the player hitting the ground. Zero collision events involving the player, or at least none with the correct normal, would indicate that the player is no longer on the ground.

I think this method may be a bit more ecs-y, and might access less of the world than `PhysicsWorld` does. However, I'm not 100% sure that the `CollisionEvent`s and `normals` work like they do in my head.

Another similar alternative would be to add a child to the player that has a sensor collider similar in shape to your `CollisionShape`. You could even spawn it dynamically on entities with `Query<&CollisionShape, Added<GroundDetection>>`. Then you could still read for `CollisionEvent`s like the first alternative, but you're still just checking for normal collisions like you do currently.

Again, this feels more ecs-y to me. I don't have an idea of what ground-detection method is standard, unfortunately. Thoughts?

Trouv (changes requested):

examples/platformer/systems.rs
ChristopherBiscardi: I moved this logic to a system+resource while I look at your other comment. I like the separation the resource adds.

examples/platformer/systems.rs
ChristopherBiscardi: I got this working via events in the latest commit. It completely gets rid of world access and is quite a bit less code. There doesn't seem to be any `normals()` data in `CollisionEvent::Stopped` though, so in the current implementation I think you could theoretically stop colliding with something that wasn't the ground and lose your jump for some length of time? I think the pattern would have to be: start colliding with ground, stop colliding with something else.

I was able to get the ghost to "lose" its ability to jump once, but couldn't replicate it again. It was when I fell through from the upper room back to the main room, I got the ghost stuck in that little area with the ladder as the level switched. I was able to climb back up the ladder and drop back down, reloading the main room and getting the jump back.

ChristopherBiscardi: as for which is more ecs-y, I suppose the events-and-resources version is. My understanding is that shape-casting is the standard in other tools like Unity which is why I tried that first, but I'm no game development expert so /shrug.

It might be worth adding some slopes to the platformer test level as well to make sure the normals method works in that case

Trouv: Yeah my bad, this definitely doesn't work as good given the getting stuck problem.

You mentioning slopes reminded me that the reason I did the normals method for an old platformer project was specifically for handling slopes. You could just do some orthogonal projection of the normal vector on Vec2::UP, and if it was above a certain value then the surface was "flat" enough to jump off of, even if it was a slope.

I think what might be happening when you get stuck is just that you're perfectly aligned with the surface you're standing on or something? No collisions are happening when you get stuck, it seems. Maybe there's some heron settings or `PhysicMaterial` that could help prevent this, but I'm not gonna put you through the ringer to try and figure that out.

Instead, I'll suggest a sort of hybrid approach between your original solution and my ecs-y stuff. It's like the second alternative in my above comment, but more detail this time. The idea is to basically do the `PhysicsWorld` solution you originally had, but with an actual entity containing the `CollisionShape` instead of creating one on the fly.
1. Make a new `GroundSensor(Entity)` component. Maybe make it `Deref` and `DerefMut` for easy access to the internal entity.
2. Make `GroundDetection` a component, add it to the `PlayerBundle`
3. Make a new system `spawn_ground_sensor` that queries for `Query<(Entity, &CollisionShape), Added<GroundDetection>>`, and spawns a child to that entity with a `GroundSensor` component pointing to the entity, and the proper `CollisionShape`, `RigidBody::Sensor`, `Transform/GlobalTransform` components to make it behave like the `CollisionShape` in your original solution.
4. Change the `ground_detection` system to find `CollisionEvent`s involving the `GroundSensor` entities and the ground, similar to the ladder logic. Then if it's colliding with the ground, update the corresponding `GroundDetection` with `Query<&mut GroundDetection>::get_mut(ground_sensor.0) = true;`. You could probably filter the collision events to ensure it's the ground by making sure their `RigidBody` isn't `RigidBody::Sensor` (rules out ladders), and to make sure they're not the internal `Entity` of the `GroundSensor`.

If you want to keep the `GroundDetection` as a resource instead of a component, I suppose that's okay, though it does change the implementation a little bit. I think I prefer it as a component as it would allow people who are making platformers based off this example to easily add ground detection to non-player entities, in case they want to make enemies who jump or something.

Thanks for your patience. I think this issue was a bit more involved than I expected it to be. Sorry about that.

What do you think?

ChristopherBiscardi: component-based ground detection sounds good to me, I'll take a crack at it. I tried using the original event-based model in a sandbox I'm working on and it failed a lot more often than even the ghost demo here.

> Thanks for your patience. I think this issue was a bit more involved than I expected it to be. Sorry about that.

No stress, I was looking for something to get involved on and this is a fine way to get into it.

ChristopherBiscardi: I updated the commits to implement the ground detection this way.

It feels to me like using CollisionEvents like this is error prone and results in hard-to-anticipate bugs (that are also hard to debug). For example, here's a bug in the current implementation where it seems like the ghost can jump if it pushes into the wall far enough that it lands on top of a wall collider segment. The jump isn't infinite, it seems to be able to second-jump in the same place every time. Perhaps I'm just not mentally modeling this correctly though.

https://user-images.githubusercontent.com/551247/166802839-094220eb-de35-4ce8-b8f4-c6b33329851d.mov

I also encountered bugs such as "skull pushes ghost into wall, jump ability is lost" during development which is hard to reproduce.

Maybe something like the intersection graph would be better: https://rapier.rs/docs/user_guides/rust/advanced_collision_detection#the-intersection-graph. Heron seems to have some support for this: https://docs.rs/heron/latest/heron/rapier_plugin/rapier2d/prelude/struct.QueryPipeline.html#method.intersections_with_shape but I haven't tried to figure out a solution using it yet. This does re-introduce the `PhysicsWorld` somewhere though since it's a `QueryPipeline` related function.

Overall, it seems like the collision-events based method still needs some work on top of what's in this commit, although it more or less currently achieves the goal of having ground detection.

Trouv: Thanks for implementing these changes! I got curious about these issues and went ahead to try and fix them on my own. I committed my changes to the `ground-detection-troubleshooting` branch, feel free to merge them into your own if you want to accept them into your PR.

Here's what I did:
1. Adjusted the shape of the ground sensor. This was to avoid climbing walls, and to make it a little more precise. This led to a hardcoded number that I didn't like: the `-4.` on line 447. Don't know why just `-half_extends.y` doesn't work.
2. Added a `GlobalTransform` to the ground sensor. It didn't have one after you removed the debug sprite which led to some weird behavior.
3. Changed the ground detection algorithm and `GroundSensor` component a bit. The idea is that, instead of setting the detection to true on `CollisionEvent::Started`, and false on `CollisionEvent::Stopped`, it keeps track of the entities that have started and haven't stopped yet. I added an `intersecting_ground_entities: HashSet<Entity>` field to the `GroundSensor`. Then, in the ground detection algorithm, I insert the colliding entity into `intersecting_ground_entities` on `CollisionEvent::Started`, and remove colliding entities from `intersecting_ground_entities` on `CollisionEvent::Stopped`. Then, I only set the detection to true if `intersecting_ground_entities.len() > 1`. This seemed to avoid a lot of the inconsistency issues, which I think were just a matter of the starts and stops getting out of sync. "Skull pushes ghost into wall, jump ability is lost" doesn't happen anymore.

There's still some weird inconsistency with whether or not ladders and dynamic rigid bodies like the enemies trip the ground sensor, but it does work a bit better now. What do you think?

ChristopherBiscardi: 1. the `-half_extends` works for the player but wouldn't that logic result in a variable ground distance detection based on the entity size? so smaller entities have to be closer to the ground to be labelled as `is_grounded == true` and larger entities could be much further away. If we're setting this up as something that could be used to apply ground detection to other entities, could this be relevant?
2. 🤦 yeah makes sense.
3. Seems more consistent with this approach. I've noticed when climbing ladders, the player can glitch into the walls by moving too far left or right, so that might account for some of the "ladders tripping the ground sensor" behavior you saw because it's actually detecting the walls as ground.

ChristopherBiscardi: I cherry picked the changes from `ground-detection-troubleshooting` into this pr

Trouv: Thanks!

examples/platformer/systems.rs
```diff
@@ -427,3 +427,63 @@ pub fn update_level_selection(
...
+                half_extends: Vec3::new(half_extends.x / 2., 2., 0.),
+                border_radius: None,
+            };
+
+            commands.entity(entity).with_children(|builder| {
+                builder
+                    .spawn()
+                    .insert(RigidBody::Sensor)
+                    .insert(detector_shape)
+                    .insert(Transform::from_xyz(0.0, -half_extends.y - 4., 0.))
```
Trouv, line 447: I think the `-4.` is necessary because of some weird parent/child scaling behavior. Try removing the `-4.` and dividing the translation by the  `GroundDection` entity's scale:
```rust
let sensor_translation = Vec3::new(0., -half_extends.y, 0.) / transform.scale;
```

examples/platformer/systems.rs
```diff
@@ -427,3 +427,63 @@ pub fn update_level_selection(
...
+pub fn ground_detection(
+    mut ground_detectors: Query<&mut GroundDetection>,
+    mut ground_sensors: Query<(Entity, &mut GroundSensor)>,
+    mut collisions: EventReader<CollisionEvent>,
+) {
+    for (entity, mut ground_sensor) in ground_sensors.iter_mut() {
+        for collision in collisions.iter() {
+            match collision {
+                CollisionEvent::Started(a, b) => {
+                    if a.rigid_body_entity() == entity {
```
Trouv, line 467: Before adding `b` to `intersecting_ground_entities`, could we also check that:
1. `b` is not the corresponding `GroundDetection` entity
2. `b` is not `RigidBody::Sensor`

Will probably need a new `Query<&RigidBody>` for the latter.

examples/platformer/systems.rs
```diff
@@ -67,7 +67,7 @@ pub fn movement(
             velocity.linear.y = (up - down) * 200.;
         }
 
-        if input.just_pressed(KeyCode::Space) {
+        if input.just_pressed(KeyCode::Space) && ground_detection.on_ground {
```
Trouv, line 70: Let's make it so you can jump out of climbing:
```suggestion
        if input.just_pressed(KeyCode::Space) && (ground_detection.on_ground || climber.climbing) {
```

Trouv (changes requested):

examples/platformer/systems.rs
ChristopherBiscardi: This fix seems to work, although tbh I don't fully understand what's causing the scale to be different here. The `transform.scale` is `Vec3(0.75, 0.75, 1.0)` and the math definitely works out to -4.6, so not surprised -4 was working.

```
[examples/platformer/systems.rs:443] sensor_translation = Vec3(
    0.0,
    -18.666666,
    0.0,
)
[examples/platformer/systems.rs:443] -half_extends.y = -14.0
```

Trouv: `bevy_ecs_ldtk` changed the scale of the entity here because, in LDtk, the size of the "Entity Instance" and its corresponding "Entity Definition" differ.

I'm still a bit confused by the whole thing as well, but I think what must be happening here is that heron's `CollisionShape`s aren't responding to a the scale of a bevy transform.

Thanks for making the change! Hope this explanation helped

examples/platformer/systems.rs
Trouv: Yeah I haven't seen it either, which is surprising to me, I definitely expected it to be an issue.

I think what's happening here is either:
1. `heron` has some specially cases w/ parent/child relationships, that children (the `GroundSensor`) can't collide with their parents (the `GroundDetection`)
2. The `CollisionEvent::Started(a, b)` *does* occur between the two, but it never gets added to the `intersecting_ground_entities` bc it occurs on the first update and the scheduling is weird

Either way, I'm okay merging this as it is since we haven't actually seen it happen. We can fix it later if we ever run into it.

Thanks for fixing point 2! Adding that comment is especially helpful

Trouv (approved):

Trouv:
Closes #44

PR #104, opened 2022-08-07
A little bit of help on the refactor/tilemap-rewrite branch

nfagerlund:
I moved the Bevy dependency to 0.8 and the bevy_ecs_tilemap one to 0.7, then started looking for whatever broke. 

~This isn't a complete upgrade and it isn't in proper working condition (the examples still need a Heron bump that isn't out yet, and also I'm still trying to untangle whatever's going on with inherited visibility), but it IS building. And all these changes seem to be things you'd have needed to do anyway, so maybe it'll save you a couple steps.~

This _seems_ to be mostly in proper working order at this point, but the examples (except `basic`) still need updates, including a Heron upgrade that I think isn't out yet. There's also some outstanding cleanup that needs done and warnings that need dealt w/, which you probably already knew about.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,6 +19,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a13739d7177fbd22bb0ed28badfff9f372f8bef46c863db4e1c6248f6b223b6e"
 
 [[package]]
+name = "adler"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+
+[[package]]
 name = "adler32"
 version = "1.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -53,7 +59,7 @@ dependencies = [
  "alsa-sys",
  "bitflags",
  "libc",
- "nix",
+ "nix 0.23.1",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


Trouv:
Yeah this is all good. I can merge it now, unless you want to make any more changes

nfagerlund:
@Trouv go for it! 🤘🏼 I’m still dinking with some stuff to see if I can figure out the visibility thing, but that wouldn’t prompt me to reconsider any of what’s already in here.

Zak0749 (changes requested):
For the from_grid_with_padding calls I think it might be better to take in the tileset_definintion padding element ,which would allow tile maps with padding, rather than using a static zero

```

TextureAtlas::from_grid_with_padding(
                tileset.clone(),
                Vec2::new(tile.w as f32, tile.h as f32),
                tileset_definition.c_wid as usize,
                tileset_definition.c_hei as usize,
                Vec2::splat(tileset_definition.spacing as f32),
                Vec2::splat(tileset_definition.padding as f32),
)
```

But other than that it looks good 👍

nfagerlund:
I think I chased down the rest of the visibility inheritance business, and got the `basic` example rendering properly! 

<details><summary>Before / after screenshots</summary>
**Before:** (main @ 8a56caf)
Screen Shot 2022-08-07 at 5 27 25 PM

**After:**
Screen Shot 2022-08-07 at 5 28 02 PM
</details>

I threw bevy-inspector-egui in there temporarily to spy on the hierarchy (when I was chasing the visibility around), and that looks to be the same too. The main difference is that, at least in debug mode, the entity sprite assets seem to pop-in somewhat later in this branch than they used to on main. 

Next step for me is to see if I can continue working on my toy project with this branch; if so, I'll probably leave the rest of the upgrade and cleanup to you! (If I get blocked there, maybe I'll add more commits. 😹)

nfagerlund:
Update: totally seems to work correctly in my toy project. 🙌🏼

Trouv:
This looks good, and the new hierarchy-related changes fixed some bugs I was dealing with before you picked this up. Thanks so much!

@Zak0749 That's a good point. I'll merge this now but I'll keep that in mind while I keep working on the update

Issue #91, opened 2022-05-27
Level spawns twice when `LdtkWorldBundle` spawns after `LevelSelection` is inserted


grace125:
The currently selected level spawns twice if `LevelSelection` already exists as a resource, and  `LdtkWorldBundle` is spawned after startup. See the following example (given some `game.ldtk`):

```rust
use bevy::prelude::*;
use bevy_ecs_ldtk::prelude::*;

fn main() {
    App::new()
        .add_state(AppState::Loading)
        .add_plugins(DefaultPlugins)
        .add_plugin(LdtkPlugin)
        .insert_resource(LevelSelection::Index(0))
        .add_system(listener)
        .add_system_set(SystemSet::on_enter(AppState::Loading).with_system(spawn_world_bundle))
        .run()
}

#[derive(Clone, Eq, PartialEq, Debug, Hash)]
enum AppState {
    Loading,
}

fn spawn_world_bundle(mut commands: Commands, asset_server: Res<AssetServer>) {
    commands.spawn_bundle(LdtkWorldBundle {
        ldtk_handle: asset_server.load("game.ldtk"),
        ..default()
    });
}

fn listener(mut level_events: EventReader<LevelEvent>) {
    level_events.iter().for_each(|l| println!("{:?}", l));
}
```

The same thing happens when `LevelSelection` is instead inserted in the `spawn_world_bundle` system. 

However, this does not happen when `LevelSelection` is first inserted later on, i.e. with a system such as:

```rust
fn insert_level_selection(input: Res<Input<KeyCode>>, mut commands: Commands) {
    if input.just_pressed(KeyCode::Space) {
        commands.insert_resource(LevelSelection::Index(0));
    }
}
```

Trouv:
Thanks for the report, I'm able to reproduce the issue.
Very strange. I'll look into this before the next release for sure.

beardedmullett:
This issue was causing me some problem with my collision detection code. The duplicate entities was messing with my collision resolution, so I did some digging.

In `fn process_ldtk_world` a new `LdtkAsset` is processed twice due to `new_ldtks: Query<&Handle<LdtkAsset>, Added<Handle<LdtkAsset>>>` query and `AssetEvent::Created` event. Removing the query and it's related code resolves the issue.

It's worth noting that it should have despawned the entities from the first pass but failed to do so because the children didn't exist on the asset.

Trouv:
[...]

PR #106, opened 2022-08-10
refactor!: update to bevy 0.8 and bevy_ecs_tilemap rewrite

Trouv:
Thanks to @nfagerlund for the help with the bevy 0.8 update and lots of troubleshooting.

There's a couple extra major changes here worth pointing out
- no more batch spawning, `bevy_ecs_tilemap` doesn't really support it in the rewrite. This will probably allow us to simplify the level spawning code, but for now it works.
- **all** tiles have `SpatialBundles` and are all children of layer entities. Before, this was only true for tiles with intgrid values and metadata. This helps with despawning levels, but it also makes the hierarchy a little less confusing. It does compromise some performance though, since potentially thousands of entities will now have components that they didn't have before.


NeoRaider:
I'm currently seeing 3 issues when updating to this branch:
- When spawning the `LdtkWorldBundle` as a child of another entity (for example to apply a `Transform`) nothing is displayed at all
- When spawning the `LdtkWorldBundle` from a non-startup system (for example a state's `on_enter`), I get various error messages about missing entities, usually leading to a crash:
```
2022-08-11T21:33:44.334253Z ERROR bevy_ecs::system::commands: Failed to 'insert or spawn' bundle of type bevy_ecs_tilemap::render::extract::ExtractedTileBundle into the following invalid entities: [9v1, 7v1, 6v1, 3v1, 4v1]
2022-08-11T21:33:44.334279Z ERROR bevy_ecs::system::commands: Failed to 'insert or spawn' bundle of type bevy_ecs_tilemap::render::extract::ExtractedTilemapBundle into the following invalid entities: [10v1, 8v1]
2022-08-11T21:33:44.334289Z ERROR bevy_ecs::system::commands: Failed to 'insert or spawn' bundle of type bevy_ecs_tilemap::render::extract::ExtractedTilemapTextureBundle into the following invalid entities: [10v1, 8v1]
thread 'Compute Task Pool (3)' panicked at 'called `Result::unwrap()` on an `Err` value: NoSuchEntity(8v1)', /home/neoraider/.cargo/registry/src/github.com-1ecc6299db9ec823/bevy_ecs_tilemap-0.7.0/src/render/prepare.rs:82:55
```

- When applying a scaling factor to the camera's `OrthographicProjection`, I'm seeing weird rendering issues at tile edges that weren't there with Bevy 0.7 (here shown with a file based on one of LDtk's example levels and a 4x scaling):

Before:
before
After:
after

Trouv:
Thanks for testing it out @NeoRaider !
> When spawning the LdtkWorldBundle as a child of another entity (for example to apply a Transform) nothing is displayed at all

I'm not able to reproduce this. Does your parent entity have a `SpatialBundle`? It needs one.

> When spawning the LdtkWorldBundle from a non-startup system (for example a state's on_enter), I get various error messages about missing entities, usually leading to a crash

Seems to be the same issue as #107. Using `bevy_ecs_tilemap` on its own this way works perfectly fine so it's definitely an issue with `bevy_ecs_ldtk`. I'm guessing it has something to do with scheduling ambiguity between the two plugins. I'll look into it now.

> When applying a scaling factor to the camera's OrthographicProjection, I'm seeing weird rendering issues at tile edges that weren't there with Bevy 0.7 (here shown with a file based on one of LDtk's example levels and a 4x scaling)

I'm also not able to reproduce this. Using the same example from LDtk, I'm changing the `OrthographicProjection::scale` and nothing's wrong on my end?

NeoRaider:
> Thanks for testing it out @NeoRaider !
> 
> > When spawning the LdtkWorldBundle as a child of another entity (for example to apply a Transform) nothing is displayed at all
> 
> I'm not able to reproduce this. Does your parent entity have a `SpatialBundle`? It needs one.

Thanks, that was the issue - I was only adding `Transform`/`GlobalTransform` instead of the whole bundle.

> 
> > When spawning the LdtkWorldBundle from a non-startup system (for example a state's on_enter), I get various error messages about missing entities, usually leading to a crash
> 
> Seems to be the same issue as #107. Using `bevy_ecs_tilemap` on its own this way works perfectly fine so it's definitely an issue with `bevy_ecs_ldtk`. I'm guessing it has something to do with scheduling ambiguity between the two plugins. I'll look into it now.
> 
> > When applying a scaling factor to the camera's OrthographicProjection, I'm seeing weird rendering issues at tile edges that weren't there with Bevy 0.7 (here shown with a file based on one of LDtk's example levels and a 4x scaling)
> 
> I'm also not able to reproduce this. Using the same example from LDtk, I'm changing the `OrthographicProjection::scale` and nothing's wrong on my end?

I'll try to reduce my code to a minimal reproducer.

Trouv:
Closes #91 
Closes #107

Trouv:
There seems to be a similar issue w/ hot reloading

NeoRaider:
My rendering issue turned out to be https://github.com/bevyengine/bevy/issues/5266. Inserting `.insert_resource(ImageSettings::default_nearest())` before `DefaultPlugins` fixed it.

Trouv:
I've promoted the Hot reloading bug to its own issue, so I'm going to merge this PR. I'll make another PR for that fix when it's ready

PR #115, opened 2022-09-01
Insert `Name` component into entities spawned by crate

Piturnah:
This pull request adds the identifier of levels and of layers to the entities spawned by `bevy_ecs_ldtk`.

## Motivation

This makes it a lot easier to see quickly what's going on in the hierarchy, for example with [bevy-inspector-egui](https://github.com/jakobhellermann/bevy-inspector-egui).

## Before

image

## After

image

### DIFF
--- a/src/level.rs
+++ b/src/level.rs
@@ -350,10 +350,12 @@ pub fn spawn_level(
                                     texture_atlases,
                                 );
 
-                                entity_commands.insert_bundle(SpatialBundle {
-                                    transform,
-                                    ..default()
-                                });
+                                entity_commands
+                                    .insert_bundle(SpatialBundle {
+                                        transform,
+                                        ..default()
+                                    })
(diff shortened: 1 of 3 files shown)


src/level.rs
```diff
@@ -350,10 +350,12 @@ pub fn spawn_level(
...
-                                entity_commands.insert_bundle(SpatialBundle {
-                                    transform,
-                                    ..default()
-                                });
+                                entity_commands
+                                    .insert_bundle(SpatialBundle {
+                                        transform,
+                                        ..default()
+                                    })
+                                    .insert(Name::new(entity_instance.identifier.to_owned()));
```
Trouv, line 358: Can you update the docs for `app::LdtkEntity::bundle_entity` at the bottom to mention that the `Name` component is inserted in addition to the `Transform` components?

Trouv (changes requested):
Thanks for the contribution!

Trouv (approved):

PR #122, opened 2022-10-12
fix: do not spawn empty ECS entity for omitted worldly entities

NeoRaider:
Skip the spawning of the entity and not only of the insertion of the components when an entity is found in the worldly_set. The empty entity doesn't seem to cause any issues, but it also doesn't do anything useful.


Trouv (approved):
Makes sense to me, good catch.

Trouv:
Clippy issue seems to be unrelated, so I'll merge

PR #127, opened 2022-11-08
refactor: move from bevy_heron to bevy_rapier2d

CashWasabi:
## Summary
The refactor consists of basically two things:
- show use of `bevy_rapier2d` in platformer example
- add `bevy_rapier2d` in cargo file and remove `heron`

For the most part I tried to not change too much of the platformer example. 
There we're some minor things though that came up whith which I had to struggle a bit.

## Notable Details
- rename `CollisionShape` -> `Collider`
- rename `Velocity.linear` -> `Velocity.linvel`
- create `SensorBundle` since there is no `RigidBody::Sensor` in `rapier2d`
- set player friction to `0.0` because the collider kept sticking to the wall when moving against it in air
- add `RapierDebugRenderPlugin` to showcase colliders with gizmos
- `pause_physics_during_load` uses RapierConfiguration now
- add `ExternalForce` to `ColliderBundle` for jumps

## Issues
I noticed that my fix for `pause_physics_during_load` might not work as excepted. Sometimes when building I have to press `R` to reload the level when I want to have the `GroundSensor` working. This might be a loading issue but I didn't find the correct solution yet.

### DIFF
--- a/examples/platformer/components.rs
+++ b/examples/platformer/components.rs
@@ -3,52 +3,55 @@ use bevy_ecs_ldtk::{prelude::*, utils::ldtk_pixel_coords_to_translation_pivoted}
 
 use std::collections::HashSet;
 
-use heron::prelude::*;
+use bevy_rapier2d::prelude::*;
 
 #[derive(Clone, Debug, Default, Bundle, LdtkIntCell)]
 pub struct ColliderBundle {
-    pub collider: CollisionShape,
+    pub collider: Collider,
     pub rigid_body: RigidBody,
     pub velocity: Velocity,
-    pub rotation_constraints: RotationConstraints,
-    pub physic_material: PhysicMaterial,
+    pub rotation_constraints: LockedAxes,
+    pub gravity_scale: GravityScale,
+    pub friction: Friction,
+    pub density: ColliderMassProperties,
+}
+
+#[derive(Clone, Debug, Default, Bundle, LdtkIntCell)]
+pub struct SensorBundle {
+    pub collider: Collider,
+    pub sensor: Sensor,
+    pub active_events: ActiveEvents,
+    pub rotation_constraints: LockedAxes,
 }
 
 impl From<EntityInstance> for ColliderBundle {
     fn from(entity_instance: EntityInstance) -> ColliderBundle {
-        let rotation_constraints = RotationConstraints::lock();
+        let rotation_constraints = LockedAxes::ROTATION_LOCKED;
 
         match entity_instance.identifier.as_ref() {
             "Player" => ColliderBundle {
-                collider: CollisionShape::Cuboid {
-                    half_extends: Vec3::new(6., 14., 0.),
-                    border_radius: None,
-                },
+                collider: Collider::cuboid(6., 14.),
                 rigid_body: RigidBody::Dynamic,
+                friction: Friction {
+                    coefficient: 0.0,
+                    combine_rule: CoefficientCombineRule::Min,
+                },
                 rotation_constraints,
                 ..Default::default()
             },
             "Mob" => ColliderBundle {
-                collider: CollisionShape::Cuboid {
-                    half_extends: Vec3::new(5., 5., 0.),
-                    border_radius: None,
-                },
+                collider: Collider::cuboid(5., 5.),
                 rigid_body: RigidBody::KinematicVelocityBased,
                 rotation_constraints,
                 ..Default::default()
             },
             "Chest" => ColliderBundle {
-                collider: CollisionShape::Cuboid {
-                    half_extends: Vec3::new(8., 8., 0.),
-                    border_radius: None,
-                },
+                collider: Collider::cuboid(8., 8.),
                 rigid_body: RigidBody::Dynamic,
                 rotation_constraints,
-                physic_material: PhysicMaterial {
-                    friction: 0.5,
-                    density: 15.0,
-                    ..Default::default()
-                },
+                gravity_scale: GravityScale(1.0),
+                friction: Friction::new(0.5),
+                density: ColliderMassProperties::Density(15.0),
                 ..Default::default()
             },
             _ => ColliderBundle::default(),
@@ -56,22 +59,21 @@ impl From<EntityInstance> for ColliderBundle {
     }
 }
 
-impl From<IntGridCell> for ColliderBundle {
-    fn from(int_grid_cell: IntGridCell) -> ColliderBundle {
-        let rotation_constraints = RotationConstraints::lock();
+impl From<IntGridCell> for SensorBundle {
+    fn from(int_grid_cell: IntGridCell) -> SensorBundle {
+        let rotation_constraints = LockedAxes::ROTATION_LOCKED;
 
+        // ladder
         if int_grid_cell.value == 2 {
-            ColliderBundle {
-                collider: CollisionShape::Cuboid {
-                    half_extends: Vec3::new(8., 8., 0.),
-                    border_radius: None,
-                },
-                rigid_body: RigidBody::Sensor,
+            SensorBundle {
+                collider: Collider::cuboid(8., 8.),
+                sensor: Sensor,
                 rotation_constraints,
+                active_events: ActiveEvents::COLLISION_EVENTS,
                 ..Default::default()
             }
         } else {
-            ColliderBundle::default()
+            SensorBundle::default()
         }
     }
 }
@@ -150,7 +152,7 @@ pub struct Climbable;
 pub struct LadderBundle {
     #[from_int_grid_cell]
     #[bundle]
-    pub collider_bundle: ColliderBundle,
+    pub sensor_bundle: SensorBundle,
     pub climbable: Climbable,
 }
 
--- a/examples/platformer/systems.rs
+++ b/examples/platformer/systems.rs
@@ -4,7 +4,7 @@ use bevy_ecs_ldtk::prelude::*;
 
 use std::collections::{HashMap, HashSet};
 
-use heron::prelude::*;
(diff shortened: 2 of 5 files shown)


examples/platformer/systems.rs
```diff
@@ -21,12 +21,12 @@ pub fn setup(mut commands: Commands, asset_server: Res<AssetServer>) {
 
 pub fn pause_physics_during_load(
```
Trouv, line 22: I think we may be able to get rid of this system now, actually. The issue it addressed with Heron doesn't seem to be a problem anymore.

examples/platformer/systems.rs
```diff
@@ -46,31 +46,42 @@ pub fn dbg_player_items(
...
             let up = if input.pressed(KeyCode::W) { 1. } else { 0. };
             let down = if input.pressed(KeyCode::S) { 1. } else { 0. };
 
-            velocity.linear.y = (up - down) * 200.;
+            velocity.linvel.y = (up - down) * 200.;
         }
 
         if input.just_pressed(KeyCode::Space) && (ground_detection.on_ground || climber.climbing) {
-            velocity.linear.y = 450.;
+            force.y = 500.;
```
Trouv, line 80: I think we probably need to increase this. On my machine, the jump is tiny, I can't even get over the skulls.

examples/platformer/systems.rs
```diff
@@ -46,31 +46,42 @@ pub fn dbg_player_items(
...
         if climber.climbing {
             let up = if input.pressed(KeyCode::W) { 1. } else { 0. };
             let down = if input.pressed(KeyCode::S) { 1. } else { 0. };
 
-            velocity.linear.y = (up - down) * 200.;
+            velocity.linvel.y = (up - down) * 200.;
         }
 
         if input.just_pressed(KeyCode::Space) && (ground_detection.on_ground || climber.climbing) {
-            velocity.linear.y = 450.;
```
Trouv, line 71: Is there a reason that we're applying an external force to it instead of a velocity?

examples/platformer/systems.rs
```diff
@@ -449,22 +441,22 @@ pub fn update_level_selection(
...
 ) {
-    for (entity, shape, transform) in detect_ground_for.iter() {
-        if let CollisionShape::Cuboid { half_extends, .. } = shape {
-            let detector_shape = CollisionShape::Cuboid {
-                half_extends: Vec3::new(half_extends.x / 2., 2., 0.),
-                border_radius: None,
-            };
+    for (entity, shape, transform) in &detect_ground_for {
+        if let Some(cuboid) = shape.as_cuboid() {
+            let Vec2 { x: hx, y: hy } = cuboid.half_extents();
```
Trouv, line 448: I'd prefer we keep variable names verbose if possible. `half_extents_x` `half_extents_y`

examples/platformer/systems.rs
```diff
@@ -449,22 +441,22 @@ pub fn update_level_selection(
...
 
-            let sensor_translation = Vec3::new(0., -half_extends.y, 0.) / transform.scale;
+            let detector_shape = Collider::cuboid(hx / 2.0, 2.);
+
+            let sensor_translation = Vec3::new(0., -hy, 0.) / transform.scale;
```
Trouv, line 452: So this ` / transform.scale` thing was a bit of a workaround for an old heron issue. I think we might be able to get rid of it now. It might be causing the ground sensor to be way lower than it should be currently

examples/platformer/systems.rs
```diff
@@ -480,31 +472,24 @@ pub fn ground_detection(
...
 ) {
-    for (entity, mut ground_sensor) in ground_sensors.iter_mut() {
+    for (entity, mut ground_sensor) in &mut ground_sensors {
         for collision in collisions.iter() {
             match collision {
-                CollisionEvent::Started(a, b) => match rigid_bodies.get(b.rigid_body_entity()) {
-                    Ok(RigidBody::Sensor) => {
-                        // don't consider sensors to be "the ground"
-                    }
+                CollisionEvent::Started(a, b, _event_flag) => match colliders.get(*b) {
```
Trouv, line 480: We may have to do something similar here that we do for the ladders where we try both orderings of `a` and `b`. By that I mean, first treat `a` as the sensor and `b` as the ground, and if it doesn't match then try `a` as the ground and `b` as the sensor. This is because collision events are only being sent once with a particular order per collision with bevy_rapier2d instead of twice with different orderings of `a` and `b`. I had to do this for my ludum dare game, which happened to be a platformer using `bevy_rapier2d`: https://github.com/Trouv/ludum-dare-51/blob/main/src/player.rs#L296

examples/platformer/systems.rs
```diff
@@ -46,31 +46,42 @@ pub fn dbg_player_items(
...
+        } else if input.pressed(KeyCode::W) || input.pressed(KeyCode::S) {
             climber.climbing = true;
         }
 
         if climber.climbing {
             let up = if input.pressed(KeyCode::W) { 1. } else { 0. };
             let down = if input.pressed(KeyCode::S) { 1. } else { 0. };
 
-            velocity.linear.y = (up - down) * 200.;
+            velocity.linvel.y = (up - down) * 200.;
```
Trouv, line 76: There seems to be some bug with players not staying on ladders when neither button is pressed like they used to. I imagine that's why you changed the `just_pressed` to `pressed` for climbing inputs. I'd like to fix this if possible

Trouv (changes requested):
Thanks so much for the contribution! I should have mentioned that I actually made a platformer with rapier for ludum dare 51, and it may be a half-decent resource for sorting out some of the issues bc the platforming turned out pretty stable there: https://github.com/Trouv/ludum-dare-51

examples/platformer/systems.rs
CashWasabi: Great code base! I used it to refactor the `ground_detection` and `detect_climb_range` function.

examples/platformer/systems.rs
CashWasabi: Removed it. :)

examples/platformer/systems.rs
CashWasabi: Not really, I got rid of it and also got rid of the `ExternalForce` component in the query.

examples/platformer/systems.rs
CashWasabi: I changed it to be 500 and it's using `velocity.linvel` directly know. This might fix the inconsistent behavior.

examples/platformer/systems.rs
CashWasabi: Deleted it.

examples/platformer/components.rs
```diff
@@ -57,21 +60,26 @@ impl From<EntityInstance> for ColliderBundle {
 }
 
 impl From<IntGridCell> for ColliderBundle {
-    fn from(int_grid_cell: IntGridCell) -> ColliderBundle {
-        let rotation_constraints = RotationConstraints::lock();
+    fn from(_: IntGridCell) -> ColliderBundle {
+        ColliderBundle::default()
+    }
+}
```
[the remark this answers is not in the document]
CashWasabi: Deleted it.

examples/platformer/components.rs
```diff
@@ -3,75 +3,79 @@ use bevy_ecs_ldtk::{prelude::*, utils::ldtk_pixel_coords_to_translation_pivoted}
...
     pub velocity: Velocity,
-    pub rotation_constraints: RotationConstraints,
-    pub physic_material: PhysicMaterial,
+    pub force: ExternalForce,
+    pub rotation_constraints: LockedAxes,
+    pub gravity_scale: GravityScale,
+    pub restitution: Restitution,
+    pub friction: Friction,
+    pub density: ColliderMassProperties,
+}
```
Trouv, line 19: I think now we can get rid of `ExternalForce` and `Restitution` no?

examples/platformer/systems.rs
```diff
@@ -258,76 +240,62 @@ pub fn spawn_wall_collision(
 
 pub fn detect_climb_range(
     mut climbers: Query<&mut Climber>,
-    climbables: Query<&Climbable>,
+    climbables: Query<Entity, (With<Collider>, With<Sensor>, With<Climbable>, Without<RigidBody>)>,
```
Trouv, line 243: I think `With<Climbable>` would suffice right? Or am I missing something

Trouv (changes requested):
This works great, just a couple minor changes left.

Also - please run `cargo fmt`, I know the CI will complain about it after I approve this.

examples/platformer/components.rs
CashWasabi: That's true. I kinda forgot to get rid of the unnecessary leftovers.

examples/platformer/systems.rs
CashWasabi: Yes, `With<Climbable>` is enough already and looks nicer, too.

CashWasabi:
I added the changes you last mentioned and ran `cargo fmt` on the codebase. I also squashed the commit since the last two commits we're just small fixes with no real information.

Trouv:
> I also squashed the commit since the last two commits we're just small fixes with no real information.

In the future, there's no need to do this for this repo since I squash commits on merge (as of 0.4) anyway. Actually, having the smaller commits before merge makes it easier to see what has changed since the last review

Trouv (approved):

janos-r:
I was working on the same update (not noticing this PR exists). In my case, I found it necessary to use `.normalize_or_zero()` instead of `.normalize()` because the first mob (spawning closest to the player), when returning, got to the exact same spot from where he was starting, creating his vector zero, having .normalize panic. I wonder why It doesn't happen here (that he doesn't return to the exact spot).

PR #136, opened 2022-11-17
fix!: adjust tile transformations for bevy_ecs_tilemap 0.8 tile position changes

Trouv:
`bevy_ecs_tilemap` 0.7 anchored tiles to the bottom-left, while 0.8 anchors tiles to the center.

For example, the (0,0) tile used to render the bottom-left pixel of the tile at the origin, but it now renders the *center* of the tile at the origin.

Since `bevy_ecs_ldtk` adds transforms to tiles, the transform logic needed to be adjusted after updating to `bevy_ecs_tilemap` 0.8. This PR performs that update, which introduces a breaking change to some utility functions

### DIFF
--- a/src/level.rs
+++ b/src/level.rs
@@ -131,7 +131,7 @@ fn spatial_bundle_for_tiles(
     layer_scale: Vec3,
 ) -> SpatialBundle {
     let mut translation =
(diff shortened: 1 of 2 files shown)


Trouv:
Thought I fixed it but I think I actually just forgot to pull the bevy_ecs_tilemap update in main woops

PR #137, opened 2022-11-17
improve ground detection

janos-r:
The ground detection was checked every tick,
and the collision event system was always iterating the sensor before the collision events. I'm not sure, but it feels like bad practice.
This PR might be just a personal preference. But it feels to me more readable and better practice/performance.
Thank you for the consideration and any response.

PS: I left the DebugRender in, as recently the grid update shifted the colliders (nicely visible with this plugin). I'm sure somebody is already considering a fix for this.

### DIFF
--- a/examples/platformer/main.rs
+++ b/examples/platformer/main.rs
@@ -15,6 +15,7 @@ fn main() {
         .add_plugins(DefaultPlugins)
         .add_plugin(LdtkPlugin)
         .add_plugin(RapierPhysicsPlugin::<NoUserData>::pixels_per_meter(100.0))
+        .add_plugin(RapierDebugRenderPlugin::default()) // draws borders around colliders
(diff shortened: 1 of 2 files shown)


Trouv:
Thanks for the contribution!

> PS: I left the DebugRender in, as recently the grid update shifted the colliders (nicely visible with this plugin). I'm sure somebody is already considering a fix for this.

That would be #136 

This seems pretty reasonable to me, but I'd prefer not to merge that `DebugRender` line into main

janos-r:
> This seems pretty reasonable to me, but I'd prefer not to merge that `DebugRender` line into main

Sorry, I was hoping it would get there before the big fix. I took it out, hope its fine now.
Thanks

Trouv (approved):
LGTM

PR #138, opened 2022-11-17
Migrated to bevy 0.9

greenfierydragon:
Does not apply to the platformer example (rapier2d) because we are waiting on https://github.com/dimforge/bevy_rapier/pull/272. When that PR is merged I will update this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,8 +15,8 @@ members = ["macros"]
 
 [dependencies]
 bevy_ecs_ldtk_macros = { version = "0.4", optional = true, path = "macros" }
-bevy_ecs_tilemap = "0.8"
-bevy = { version = "0.8", default-features = false, features = ["bevy_sprite"] }
+bevy_ecs_tilemap = "0.9"
+bevy = { version = "0.9", default-features = false, features = ["bevy_sprite"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 regex = "1"
@@ -25,7 +25,7 @@ anyhow = "1.0"
 thiserror = "1.0"
 
 [dev-dependencies]
-bevy = "0.8"
+bevy = "0.9"
 bevy_rapier2d = { version = "0.18" }
 rand = "0.8"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ thiserror = "1.0"
 
 [dev-dependencies]
 bevy = "0.9"
-bevy_rapier2d = { version = "0.18" }
+bevy_rapier2d = { git = "https://github.com/devil-ira/bevy_rapier.git", branch = "bevy-0.9" }
 rand = "0.8"
 
 [features]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ thiserror = "1.0"
 
 [dev-dependencies]
 bevy = "0.9"
-bevy_rapier2d = { git = "https://github.com/devil-ira/bevy_rapier.git", branch = "bevy-0.9" }
+bevy_rapier2d = "0.19"
 rand = "0.8"
 
 [features]
--- a/src/resources.rs
+++ b/src/resources.rs
@@ -2,8 +2,7 @@
 
 use crate::ldtk::Level;
 
-#[allow(unused_imports)]
-use bevy::prelude::{CoreStage, GlobalTransform, Resource};
+use bevy::prelude::*;
 
 #[allow(unused_imports)]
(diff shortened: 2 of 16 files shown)


src/resources.rs
```diff
@@ -3,7 +3,7 @@
 use crate::ldtk::Level;
 
 #[allow(unused_imports)]
-use bevy::prelude::{CoreStage, GlobalTransform};
+use bevy::prelude::{CoreStage, GlobalTransform, Resource};
```
Trouv, line 6: This `#[allow(unused_imports)]` import was originally intended for imports that aren't used by in code but *are* referenced by the docs. Little did I know that rust-doc doesn't actually link those up properly anyway. Since we now actually use an item from `prelude` in this module, can we update it to be `use bevy::prelude::*;` and remove the `#[allow(unused_imports)]`?

src/utils.rs
```diff
@@ -280,13 +280,13 @@ pub fn sprite_sheet_bundle_from_entity_info(
...
-            texture_atlas: texture_atlases.add(TextureAtlas::from_grid_with_padding(
+            texture_atlas: texture_atlases.add(TextureAtlas::from_grid(
                 tileset.clone(),
                 Vec2::new(tile.w as f32, tile.h as f32),
                 tileset_definition.c_wid as usize,
                 tileset_definition.c_hei as usize,
-                Vec2::splat(tileset_definition.spacing as f32),
-                Vec2::splat(tileset_definition.padding as f32),
+                Some(Vec2::splat(tileset_definition.spacing as f32)),
+                Some(Vec2::splat(tileset_definition.padding as f32)),
```
Trouv, line 289: This change to `TextureAtlas` needs to be updated in at least one more place, line 232 of `macros/src/ldtk_entity.rs`. Feel free to just wrap `Some` around the padding and offset values without changing the macro itself in that case.

Trouv (changes requested):
Thanks for the contribution!

Trouv:
> When that PR is merged I will update this.

Feel free to depend on that PR in main. I'm also down to update that example in a separate PR if it's a lot of changes.

johanhelsing:
Not sure if it's this PR, or an existing bevy_ecs_ldtk issue, but I have problems with the background and tiles not lining up:

image

The white box is the background for the level with the sunnyland grass textures, so either the positions of the background or the tiles is off.

image

Trouv:
> I have problems with the background and tiles not lining up:

This should be fixed by #136, this PR hasn't pulled from main since that was merged

src/resources.rs
greenfierydragon: Am I correct in saying this to be done across all files?

Trouv: Nah just this one is fine since this is the only one being updated in this PR. The others are a bit unrelated to the purpose of this PR and can be changed later

src/utils.rs
greenfierydragon: Resolves in 222a46f

greenfierydragon:
> I think
> 
> ```rust
> app.add_plugin(bevy_ecs_tilemap::TilemapPlugin)
> ```
> 
> should probably be removed from `LdtkPlugin::build`. If more than one plugins does this, the user will get panics like:
> 
> ```
> thread 'main' panicked at 'Error adding plugin bevy_ecs_tilemap::TilemapPlugin: : plugin was already added in application', C:\Users\Johan\.cargo\registry\src\github.com-1ecc6299db9ec823\bevy_app-0.9.0\src\app.rs:841:63
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```
> 
> Alternatively, put it behind a `app.is_plugin_added` check.

Fixed in 90092d8

greenfierydragon:
Rebased based of upstream

Cargo.toml
```diff
@@ -25,8 +25,8 @@ anyhow = "1.0"
 thiserror = "1.0"
 
 [dev-dependencies]
-bevy = "0.8"
-bevy_rapier2d = { version = "0.18" }
+bevy = "0.9"
+bevy_rapier2d = { git = "https://github.com/devil-ira/bevy_rapier.git", branch = "bevy-0.9" }
```
johanhelsing, line 29: v0.19.0 is out now :)

johanhelsing (commented):
Alignment issues gone :)

No other issues

Cargo.toml
greenfierydragon: Should be there now!

Trouv:
Pushing a formatting change...

Trouv (approved):
LGTM. Thanks for your hard work @greenfierydragon and @johanhelsing - I really appreciate it

PR #140, opened 2022-11-23
Fix levels having incorrect `GlobalTransform` the first frame

johanhelsing:
The transform propagation system runs in post_update, so spawn the levels before `PostUpdate` so they have time to propagate, but after ProcessApi, so they'll be ready the next frame (assuming the LdtkWorldBundle will be spawned during the `Update`) stage.

I ran into this bug because in my game I scale the levels, and collision detection is therefore way off during the first frame sending my player character flying.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -147,6 +147,7 @@ mod plugin {
         /// As a result, you can expect minimal frame delay when updating these in
         /// [CoreStage::Update].
         ProcessApi,
+        SpawnLevelsStage,
     }
 
     /// Adds the default systems, assets, and resources used by `bevy_ecs_ldtk`.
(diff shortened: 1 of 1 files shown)


Trouv:
I have to give this some thought. I'd prefer not to add another stage if possible, but I can't think of a way to off the top of my head. One reason I wanted to put level spawning before `Update` was so that users could flesh out entities using the "blueprint pattern" (querying for `Added<EntityInstance>` or `Added<MarkerComponent>`) during `Update` and not have any frame delay. Though, this seems like a reasonable concern too.

andriyDev:
Hi, I believe I just stumbled upon this myself all the way in 0.15. I "fixed" this by adding:

```rust
.add_systems(
  RunFixedMainLoop,
  (sync_simple_transforms, propagate_transforms)
    .in_set(RunFixedMainLoopSystem::BeforeFixedMainLoop),
)
```

This means the ldtk stuff spawned in PreUpdate will have their transform propagated just before avian gets a hold of it. Obviously this isn't really a desirable solution though since it adds in a whole extra transform propagation when most frames won't need/want it.

What if we just put the spawning stuff in `PostUpdate` before `TransformSystem::TransformPropagate`? That just means the level will be ready on the subsequent frame. WDYT @Trouv?

PR #142, opened 2022-11-28
docs: Explain that sprite_bundle should not be used with tilemap editor visuals

johanhelsing:
I was wondering whether I hit a bug before I scrolled down and found sprite_sheet_bundle


Trouv (approved):
Thanks!

PR #144, opened 2022-12-03
feat: Add LevelSet::from_iid

johanhelsing:
If you want to spawn multiple levels at once and handle loading unloading yourself (not using global resources), it's useful to be able to do things like this:

```rust
commands.spawn(LdtkWorldBundle {
    ldtk_handle: asset_server.load("levels.ldtk"),
    level_set: LevelSet::from_iid("58e02ac0-5110-11ed-b070-cdccc74eaccd"),
    ..default()
})
```

### DIFF
--- a/src/components.rs
+++ b/src/components.rs
@@ -45,6 +45,14 @@ pub struct LevelSet {
     pub iids: HashSet<String>,
 }
 
+impl LevelSet {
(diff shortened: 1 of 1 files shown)


Trouv (approved):
Thanks, this is great

PR #146, opened 2022-12-11
feat: Register types `GridCoords` and `LayerMetadata`

johanhelsing:
Types need to be registered for `bevy-inspector-egui` (and other tools) to know about them. Do it in the plugin so it's convenient for users.


Trouv (approved):
Makes sense, thanks for the contribution.

PR #147, opened 2022-12-18
example/improve-spawn_wall_collision

janos-r:
I thought that this part of the collision builder was unnecessarily complicated and I didn't like that it is creating new HashMaps on every row. I tried to measure the the difference with std::time::Instant, but turns out the difference is in microseconds, not much improvement in time. But I still decided to submit this PR because I think it is a slight improvement.
What do you think?


Trouv (approved):
Seems pretty reasonable to me. Even though it's not much of a perf improvement, it also looks a little bit cleaner, so I'm down to merge. Thanks for the contribution.

PR #149, opened 2022-12-28
refactor: change #[from_entity_instance] to use references

NeoRaider:
EntityInstance is a rather large struct, and cloning it requires allocating for several fields. Despite this fact, each use of the #[from_entity_instance] attribute was cloning the struct once, wasting CPU time for no good reason - especially for Bundles that contain several #[from_entity_instance] fields.

Change the attribute to work on From<&EntityInstance> instead of From<EntityInstance>, avoiding the implicit clone in most cases. The clone is still necessary when a plain EntityInstance field exists in a struct deriving LdtkEntity, which is handled by having EntityInstance implement From<&EntityInstance>.

While being a breaking change, fixing up users should be trivial - in most cases no owned EntityInstance is needed at all, and where it is an explicit clone can be inserted.

### DIFF
--- a/examples/platformer/components.rs
+++ b/examples/platformer/components.rs
@@ -24,8 +24,8 @@ pub struct SensorBundle {
     pub rotation_constraints: LockedAxes,
 }
 
-impl From<EntityInstance> for ColliderBundle {
-    fn from(entity_instance: EntityInstance) -> ColliderBundle {
+impl From<&EntityInstance> for ColliderBundle {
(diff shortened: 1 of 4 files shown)


Trouv:
This makes perfect sense, thanks for the contribution. I bet there's several places in the repo where I was a bit too clone-happy where I could have borrowed. This is definitely a breaking change though, so I'll mark it as such in the squashed commit.

Issue #47, opened 2022-01-23
System-based LdtkEntity and LdtkIntCell


Trouv:
There are two main drawbacks for using `LdtkEntity` and `LdtkIntCell`
1. For mildly complex cases, implementing these traits manually is super verbose (many method arguments)
2. For more complex cases, the parameters may not give you enough access to the world

Basically, these two traits are very opinionated about what the user might need when they are spawning bundles from the Entity and IntGrid layers. This can be mildly annoying. I think an ideal solution would be to use bevy systems for spawning. That way, users can define exactly what they need from the world, and can feel free to not ask for what they don't need, reducing boilerplate.

Not sure how the exact implementation of this would work (probably using some dark, forbidden bevy traits like `FunctionSystem`). I'd like the derive traits to be at least as useful as they are now. I'm okay with the design of the macros changing somewhat as long as it's an improvement.

I don't have much of an opinion on whether `LdtkEntity` and `LdtkIntCell` should even remain traits with this change. I feel like it would probably be pushing rust's generic features to the limits to have generic trait-method parameters, but idk. I may slightly prefer that they remain traits just to make typing easier as they are passed around the plugin.

This change may result in the level-spawning system becoming exclusive. This is not ideal, but I can't really imagine an alternative. If somehow the level-spawning system can be generic over whatever the user defines in their spawning systems that they register to the app, that would probably be ideal, but it sounds impossible.

Also, not sure how exactly to pass the LDtk data to these systems (`EntityInstance`, `LayerInstance`, `IntGridCell`). Maybe an `In` parameter? A custom system param?

Simre1:
What do you think about using bevy events for this?

Whenever a level/layer/entity/intcell is added, an event with the relevant information is fired. Systems listen to these events and do the spawning.

PR #156, opened 2023-01-27
Spawn hook

geieredgar:
Hello, first I want to thank you for this useful crate. I am currently using it for a hobby project.

This is my suggestion for a possible solution of #47. It is implemented without breaking changes.

The first commit adds a new `SpawnHook` trait.  It defines an associated `Param` type which allows accessing more from the world when executing the `process_ldtk_levels` system. The `DefaultSpawnHook` type implements `SpawnHook` and does the spawning in the same way as before.

The second commit adds a new `spawn` module that adds another implementation of the `SpawnHook` trait, the `Registry`. The registry allows hooking up individual functions that implement `SystemParamFunction` and using another trait called `SpawnFunction` that can be used with types that implement `From<SpawnFunction::Input<'a>>`.

I would add documentation and more examples if there is an interest in these additions.

### DIFF
--- a/examples/custom_spawn.rs
+++ b/examples/custom_spawn.rs
@@ -1,0 +1,73 @@
+use bevy::prelude::*;
+use bevy_ecs_ldtk::{prelude::*, spawn::*};
+
+fn main() {
+    App::new()
+        .add_plugins(
+            DefaultPlugins.set(ImagePlugin::default_nearest()), // prevents blurry sprites
+        )
+        .add_plugin(LdtkPluginWithSpawnHook(registy))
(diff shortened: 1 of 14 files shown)


Trouv:
Thanks for all the contributions yesterday! This is very interesting. Actually, this particular issue is one I've been thinking about a lot lately, so this couldn't have come at a better time. Defining custom `Params` for the `process_ldtk_levels` system is especially interesting to me. I'll have a closer look at this some time in the future, and I can't promise right now that this is the solution that I will adopt, but it has some interesting ideas that I hadn't considered before.

PR #157, opened 2023-01-27
Add `with` attribute for LdtkIntCell derive macro

### LINKED ISSUE #135: Add `with` attribute for LdtkIntCell derive macro
#128 added a `with` attribute to the LdtkEntity derive macro. It's a good idea, and we should do something similar for the int cell macro.

geieredgar:
Fixes #135.


Trouv (approved):
LGTM, thanks again

PR #159, opened 2023-01-27
Headless mode

geieredgar:
This allows running in headless mode by disabling the added `render` feature.


Trouv (approved):
This doesn't really hurt anything unless we release the next version before bevy_ecs_tilemap does (headless mode doesn't compile in 0.9), which is unlikely

Trouv:
Is it bad for headless mode users that, even though the tilemaps are disabled in headless mode, the sprites spawed for `LdtkEntity`s still spawn?

PR #161, opened 2023-01-27
Allow opting out of SpriteSheetBundle grid construction

### LINKED ISSUE #105: Rectangular tile index incorrect when not a multiple of tile width/height
This seemingly occurs when a [rectangle of tiles](https://github.com/deepnight/ldtk/issues/533) is chosen for an entity, and the offset of this rectangle is not a multiple of the tile width or height.

Most likely a floor-vs-ceil issue in [`#[sprite_sheet_bundle]`](https://github.com/Trouv/bevy_ecs_ldtk/blob/main/src/utils.rs#L287-L289), but I am not super familiar with the intricacies involved here.

More than happy to provide a reproduction case if required.

---

Compare this LDtk screenshot:

Screenshot 2022-08-07 at 10 14 01

With the `bevy_ecs_ldtk` rendered result (both entities on the right are off by one column and row respectively):

Screenshot 2022-08-07 at 10 14 15

### LINKED ISSUE #151: Sprite selection from multiple tiles loads the wrong tiles
bevy_ecs_ldtk: 0.5
ldtk editor: 1.2.3

Short summary:
When choosing a 2x2 tile sprite from the sheet, the library loads the correct tiles only under the assumption, that all the sprites in this sheet are made in the same size.
But such a sheet would be effectively always selected as 1x1. In other words, choosing a sprite from several tiles is currently broken.

Example from the current platformer:
1) Choose a 2x2 selection for the chest entity. But don't select from the first or third line. Select a 2x2 starting on the second, third or any other odd line. Note: The pritesheet used here has the first line empty.
image

2) save and load the platformer - the sprites loaded will be shifted one line higher.
image

This is not an issue when selecting only 1x1 tiles, but it makes it impossible to use a sheet set like this:
TilesetHouse
That is how I encountered this issue. When loading the torii gate, you shift one line up. The same goes for most of the houses that are not starting on the first line.

Thank you for your consideration.

geieredgar:
This is an attempt to fix #105 and fix #151.

This adds an optional `no_grid` argument to the `#[sprite_sheet_bundle]` attribute. Instead of using `TextureAtlas::from_grid`, a `TextureAtlas` with a single sprite will be used.  

Instead of using `#[sprite_sheet_bundle(no_grid)]`, a field of type `NoGridSpriteSheetBundle` annotated with the `#[ldtk_entity]` attribute can be used.


src/components.rs
```diff
@@ -371,3 +371,9 @@ pub struct LdtkWorldBundle {
     pub visibility: Visibility,
     pub computed_visibility: ComputedVisibility,
 }
+
+/// [Bundle] for creating a SpriteSheetBundle without using a tileset grid. Intended for use with [#[ldtk_entity]](LdtkEntity#ldtk_entity).
+#[derive(Bundle)]
+pub struct NoGridSpriteSheetBundle {
```
Trouv, line 377: I think the addition of this bundle is a bit unnecessary. I'm happy with just the `#[sprite_sheet_bundle(no_grid)]` attribute

src/app/ldtk_entity.rs
```diff
@@ -104,6 +104,8 @@ use std::{collections::HashMap, marker::PhantomData};
 /// Similar to using [TextureAtlas::from_grid()].
 /// - `#[sprite_sheet_bundle]` will create the field using information from the LDtk Editor visual,
 /// if it has one.
+/// /// - `#[sprite_sheet_bundle(no_grid)]` will create the field using information from the LDtk Editor visual,
```
Trouv, line 107:
```suggestion
/// - `#[sprite_sheet_bundle(no_grid)]` will create the field using information from the LDtk Editor visual,
```

src/app/ldtk_entity.rs
```diff
@@ -104,6 +104,8 @@ use std::{collections::HashMap, marker::PhantomData};
...
+/// if it has one, but without using a grid. Instead a single texture will be used.
```
Trouv, line 108: Could we add something along the lines of "This may be useful if the LDtk entity's visual uses a rectangle of tiles from its tileset, but will prevent using the generated TextureAtlas for animation purposes."

Trouv (changes requested):
Thanks for the contribution, I see now that it would be difficult to support these offset rectangles of tiles with grids.

src/components.rs
geieredgar: Ok, I removed everything related to `NoGridSpriteSheetBundle`.

src/app/ldtk_entity.rs
geieredgar: Done.

src/app/ldtk_entity.rs
geieredgar: Fixed.

examples/basic.rs
```diff
@@ -36,3 +37,10 @@ pub struct MyBundle {
     #[bundle]
     sprite_bundle: SpriteSheetBundle,
 }
+
+#[derive(Bundle, LdtkEntity)]
+pub struct NoGridEntity {
+    #[sprite_sheet_bundle(no_grid)]
+    #[bundle]
+    sprite_bundle: SpriteSheetBundle,
+}
```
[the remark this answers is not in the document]
geieredgar: OK, I've restored the basic example and updated the platformer example instead.

Trouv (approved):
LGTM, thanks again

PR #162, opened 2023-01-27
Respect pivot in tile layers

### LINKED ISSUE #152: Rules based tiles don't use pivot
LDtk: 1.2.4
bevy_ecs_ldtk: 0.5
bevy: 0.9

Issue:
The pivot used in LDtk editor for autotiled sprites is not respected in the library, which is visible when used with sprites more than 1 tile big. The pivot used in the library is equal as if all pivot in rules-based tiles was set to the bottom-right corner.

Example:
scrnshot from the current unchanged platformer example, inside the LDtk editor:
Screenshot from 2023-01-11 18-28-59
The rule is always based on the center grid.
Screenshot from 2023-01-11 18-25-08
It is shown on the map with a violet frame. The pivot says that the picture of this gray shelf should be in the middle of that frame, but this pivot is not respected in the actual loaded game with the library:
Screenshot from 2023-01-11 18-29-42
The shelf is drawn one grid to the left from where it should be.

In this situation it is not a big issue in the case of platformer, where autotiles are used as random scatters, but this is a big issue when 
a) it is either important to be precisely aligned with the surroundings
b) when the concerned intGrid has other effects, like collision. If this shelf had collision, the player would bump into the empty space next to the shelf.

Example from my own project where the auto tiles are drawing trees and stumps.
Screenshot from 2023-01-11 18-31-17
Not only are they visually not in the right alignment with the fences (the fences just one tile big, so this issue doesn't apply to them), but when used with collision, the player would collide with the grid (as desired) but the tree sprite is somewhere else.

I can workaround this by redoing the auto-tiles as objects, but this should be fixed if somebody can find the cause, or introduce the feature to the library. It would be impossible to properly use auto-tiles bigger than one tile.

I hope my explanation was clear.
Thanks to anybody working on this amazing library and anybody helping out!

geieredgar:
This should fix #152.

Tiles which are not aligned to the grid are put into own layers, one layer for each encountered alignment offset. 
They also only span the part of the grid in which tiles are found. 
Another separate layer is added that contains the entities for the int grid cells.


Trouv:
Finally starting to look at this. It's kinda scary changing stuff like this. Here's an example why - check out the platformer example, the tops of the ladders don't seem to be acting as ladders. Not sure why yet

geieredgar:
I've found the bug, it was this line:
```rust
max: IVec2::new(layer_instance.c_wid, layer_instance.c_hei),
```
It should have been
```rust
max: IVec2::new(layer_instance.c_wid - 1, layer_instance.c_hei - 1),
```

During debugging I've tried out several things and am now wondering if it would be better to have one entity that corresponds to the layer (like before, but without the `TileMapBundle`, because we may need several of those).
Then the tile-maps for each sub-layer would be a children of the layer entity, and the tile entities would also be children of the layer entity instead of the tile map entity. This would restore the previous behavior that all tiles of a layer are children of the same entity. 

I am also wondering if we should add spatial bundles only to tile entities created from the IntGridCell, because all other entities should probably only be used for rendering.

Trouv:
> During debugging I've tried out several things and am now wondering if it would be better to have one entity that corresponds to the layer (like before, but without the TileMapBundle, because we may need several of those).
Then the tile-maps for each sub-layer would be a children of the layer entity, and the tile entities would also be children of the layer entity instead of the tile map entity. This would restore the previous behavior that all tiles of a layer are children of the same entity.

This makes sense to me - though I'm not sure off the top of my head if `bevy_ecs_tilemap` requires tile entities to be children of tilemaps?

If that hierarchy is required, I think that's fine. That would be a breaking change for many users, but it sounds like a logical hierarchy to me and it's not like we're a stable library.

> I am also wondering if we should add spatial bundles only to tile entities created from the IntGridCell, because all other entities should probably only be used for rendering.

`IntGridCell`s **and** tiles w/ tile metadata. I think giving it to EVERY tile is a holdover from when we used batch-spawning, since they all had to have the same bundle. So, unless there's some other use case for that I'm forgetting, that makes sense to me too.

Trouv (changes requested):
Finally got around to reviewing this again - thanks for your patience

PR #168, opened 2023-03-06
Update to Bevy `0.10`

geieredgar:
Currently blocked by

- [ ] https://github.com/StarArawn/bevy_ecs_tilemap/pull/390
- [x] https://github.com/dimforge/bevy_rapier/pull/324
- [x] https://github.com/dimforge/nalgebra/pull/1213
- [x] Fix system race conditions with `bevy_rapier` in `platformer` example

### DIFF
--- a/examples/platformer/main.rs
+++ b/examples/platformer/main.rs
@@ -2,7 +2,7 @@
 // Please run with `--release`.
 
 use bevy::prelude::*;
-use bevy_ecs_ldtk::{prelude::*, LdtkSystemSet};
+use bevy_ecs_ldtk::prelude::*;
 
 use bevy_rapier2d::prelude::*;
 
--- a/examples/platformer/main.rs
+++ b/examples/platformer/main.rs
@@ -2,7 +2,7 @@
 // Please run with `--release`.
 
 use bevy::prelude::*;
-use bevy_ecs_ldtk::prelude::*;
+use bevy_ecs_ldtk::{prelude::*, LdtkSystemSet};
 
 use bevy_rapier2d::prelude::*;
 
@@ -14,6 +14,7 @@ fn main() {
         .add_plugins(DefaultPlugins.set(ImagePlugin::default_nearest()))
         .add_plugin(LdtkPlugin)
         .add_plugin(RapierPhysicsPlugin::<NoUserData>::pixels_per_meter(100.0))
+        .configure_set(LdtkSystemSet::ProcessApi.before(PhysicsSet::SyncBackend))
         .insert_resource(RapierConfiguration {
             gravity: Vec2::new(0.0, -2000.0),
             ..Default::default()
--- a/examples/platformer/main.rs
+++ b/examples/platformer/main.rs
@@ -14,6 +14,7 @@ fn main() {
         .add_plugins(DefaultPlugins.set(ImagePlugin::default_nearest()))
         .add_plugin(LdtkPlugin)
         .add_plugin(RapierPhysicsPlugin::<NoUserData>::pixels_per_meter(100.0))
+        // Required to prevent race conditions between bevy_ecs_ldtk's and bevy_rapier's systems
         .configure_set(LdtkSystemSet::ProcessApi.before(PhysicsSet::SyncBackend))
         .insert_resource(RapierConfiguration {
             gravity: Vec2::new(0.0, -2000.0),
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -157,8 +157,9 @@ mod plugin {
     }
 
     #[derive(Copy, Clone, Eq, PartialEq, Debug, Hash, SystemSet)]
-    enum InternalSystemSet {
+    enum ProcessApiSet {
         PreClean,
+        Clean,
     }
 
     /// Adds the default systems, assets, and resources used by `bevy_ecs_ldtk`.
@@ -179,6 +180,11 @@ mod plugin {
                     .after(CoreSet::UpdateFlush)
                     .before(CoreSet::PostUpdate),
             )
+            .configure_sets(
+                (ProcessApiSet::PreClean, ProcessApiSet::Clean)
+                    .chain()
+                    .in_base_set(LdtkSystemSet::ProcessApi),
+            )
             .init_non_send_resource::<app::LdtkEntityMap>()
             .init_non_send_resource::<app::LdtkIntCellMap>()
             .init_resource::<resources::LdtkSettings>()
@@ -191,22 +197,16 @@ mod plugin {
                 (systems::process_ldtk_assets, systems::process_ldtk_levels)
                     .in_base_set(CoreSet::PreUpdate),
             )
-            .add_system(
-                systems::worldly_adoption
-                    .in_base_set(LdtkSystemSet::ProcessApi)
-                    .in_set(InternalSystemSet::PreClean),
-            )
+            .add_system(systems::worldly_adoption.in_set(ProcessApiSet::PreClean))
             .add_systems(
                 (systems::apply_level_selection, systems::apply_level_set)
                     .chain()
-                    .in_base_set(LdtkSystemSet::ProcessApi)
-                    .in_set(InternalSystemSet::PreClean),
+                    .in_set(ProcessApiSet::PreClean),
             )
             .add_systems(
                 (apply_system_buffers, systems::clean_respawn_entities)
                     .chain()
-                    .in_base_set(LdtkSystemSet::ProcessApi)
-                    .after(InternalSystemSet::PreClean),
+                    .in_set(ProcessApiSet::Clean),
             )
             .add_system(
                 systems::detect_level_spawned_events
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -138,11 +138,11 @@ mod plugin {
 
     use super::*;
 
-    /// [StageLabel] for stages added by the plugin.
+    /// Base [SystemSet]s for systems added by the plugin.
     #[derive(Copy, Clone, Eq, PartialEq, Debug, Hash, SystemSet)]
     #[system_set(base)]
     pub enum LdtkSystemSet {
-        /// Occurs immediately after [CoreStage::Update].
+        /// Occurs immediately after [CoreSet::UpdateFlush].
         ///
         /// Used for systems that process components and resources provided by this plugin's API.
         /// In particular, this stage processes..
@@ -152,9 +152,13 @@ mod plugin {
         /// - [components::Respawn]
         ///
         /// As a result, you can expect minimal frame delay when updating these in
-        /// [CoreStage::Update].
+        /// [CoreSet::Update].
         ProcessApi,
-        PostProcessApi,
+    }
+
+    #[derive(Copy, Clone, Eq, PartialEq, Debug, Hash, SystemSet)]
+    enum InternalSystemSet {
+        PreClean,
     }
 
     /// Adds the default systems, assets, and resources used by `bevy_ecs_ldtk`.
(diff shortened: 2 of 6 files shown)


Trouv (changes requested):
Thanks for your hard work updating this and `bevy_ecs_tilemap`. It really made my already-busy week so much easier

src/lib.rs
```diff
@@ -178,10 +170,11 @@ mod plugin {
...
             }
 
-            app.add_stage_after(
-                CoreStage::Update,
-                LdtkStage::ProcessApi,
-                SystemStage::parallel(),
+            app.configure_sets(
+                (LdtkSystemSet::ProcessApi, LdtkSystemSet::PostProcessApi)
+                    .chain()
+                    .after(CoreSet::Update)
```
Trouv, line 176: I think technically this should come after `CoreSet::UpdateFlush`, since that's where the post-`Update` `apply_system_buffers` lives. `apply_level_selection` and, less importantly, `worldly_adoption` both need this buffer to minimize update-delay for users

Trouv (changes requested):

src/lib.rs
```diff
@@ -138,19 +138,10 @@ mod plugin {
...
-    #[derive(Copy, Clone, Eq, PartialEq, Debug, Hash, SystemLabel)]
-    pub enum LdtkSystemLabel {
-        ProcessAssets,
-        LevelSelection,
-        LevelSet,
-        LevelSpawning,
-        Other,
-    }
-
     /// [StageLabel] for stages added by the plugin.
```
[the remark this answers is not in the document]
geieredgar: Done.

src/lib.rs
```diff
@@ -191,37 +184,21 @@ mod plugin {
...
-                systems::clean_respawn_entities.at_end(),
+            .add_system(systems::worldly_adoption.in_base_set(LdtkSystemSet::ProcessApi))
+            .add_systems(
+                (systems::apply_level_selection, systems::apply_level_set)
+                    .chain()
+                    .in_base_set(LdtkSystemSet::ProcessApi),
             )
-            .add_system_to_stage(
-                CoreStage::PostUpdate,
+            .add_system(systems::clean_respawn_entities.in_base_set(LdtkSystemSet::PostProcessApi))
```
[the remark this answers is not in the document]
geieredgar: You are right, there was an `apply_system_buffers` missing.

src/lib.rs
geieredgar: Done.

src/lib.rs
```diff
@@ -163,6 +154,7 @@ mod plugin {
         /// As a result, you can expect minimal frame delay when updating these in
         /// [CoreStage::Update].
         ProcessApi,
+        PostProcessApi,
```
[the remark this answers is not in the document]
geieredgar: This sounds good, so I implemented the nested set idea. I've added a new non-base system set type `InternalSystemSet`. I've decided to make it non-pub, because I think this type mainly encodes implementation details, so leaving it out of the public API allows non-breaking changes. I would like to hear your feedback about that.

geieredgar:
I've addressed all of the requested changes. Unfortunately, I've encountered what I believe is some race condition between systems of `bevy_ecs_ldtk` and `bevy_rapier` in the `platformer` example. I've experienced them before, but hoped this was an issue with the then WIP `bevy_rapier` update. I will investigate them later today.

src/lib.rs
```diff
@@ -191,37 +187,31 @@ mod plugin {
...
+                (systems::process_ldtk_assets, systems::process_ldtk_levels)
+                    .in_base_set(CoreSet::PreUpdate),
             )
-            .add_system_to_stage(
-                LdtkStage::ProcessApi,
-                systems::apply_level_selection.label(LdtkSystemLabel::LevelSelection),
+            .add_system(
+                systems::worldly_adoption
+                    .in_base_set(LdtkSystemSet::ProcessApi)
+                    .in_set(InternalSystemSet::PreClean),
```
Trouv, line 197: I think you can just `.configure_set(InternalSystemSet::PreClean.in_base_set(LdtkSystemSet::ProcessApi))`, and then you don't need to specify `.in_base_set(LdtkSystemSet::ProcessApi)` on the systems in `PreClean`

src/lib.rs
```diff
@@ -161,10 +152,15 @@ mod plugin {
         /// - [components::Respawn]
         ///
         /// As a result, you can expect minimal frame delay when updating these in
-        /// [CoreStage::Update].
+        /// [CoreSet::Update].
         ProcessApi,
     }
 
+    #[derive(Copy, Clone, Eq, PartialEq, Debug, Hash, SystemSet)]
+    enum InternalSystemSet {
```
Trouv, line 160: Do you think we could name this something like `ProcessApiSet`? Since it's only going to be used in `ProcessApi`

src/lib.rs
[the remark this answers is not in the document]
Trouv: Yeah keeping it private is fine. If somebody feels like they need it they can file an issue in the future, but I don't expect that to happen. Actually, I feel like this plugin has too many public items as it is

src/lib.rs
geieredgar: Ok, I also added a `Clean` set and used that to order the systems, so that we can do `.in_set(ProcessApiSet::Clean)` instead of `.in_base_set(LdtkSystemSet::ProcessApi).after(ProcessApiSet::Clean)`.

src/lib.rs
geieredgar: Done.

geieredgar:
Here are my investigation results about the race condition in the `platformer` example:

Without any explicit ordering between `LdtkSystemSet::ProcessApi` and `PhysicsSet::SyncBackend`, the platformer example quite often crashes at startup with the following error:
```
thread 'main' panicked at 'error[B0003]: Could not insert a bundle (of type `bevy_rapier2d::dynamics::rigid_body::RapierRigidBodyHandle`) for entity 3814v0 because it doesn't exist in this World.
```

The system that issues the command to insert the bundle is `init_rigid_bodies` which is part of the `PhysicsSet::SyncBackend` set.

Even if the example does not crash, the player is sometimes spawned at the wrong position, see

Wrong spawn position

Ordering `PhysicsSet::SyncBackend` before `LdtkSystemSet::ProcessApi` didn't crash anymore, but the wrong spawn position did also happen occasionally (like about 1 of 30 times). Sometimes the player didn't spawn at all (maybe it was only invisible/occluded by the level), see

Player didn't spawn

Therefore, I changed the `platformer` example to explicitly schedule `PhysicsSet::SyncBackend` after `LdtkSystemSet::ProcessApi`. With this I didn't encounter any crashes or wrong spawn positions, but this doesn't prove it cannot happen.

### Side note

As a side note, I've run the `platformer` example in release mode and looked at the first few frames using renderdoc. The frames looked the same between all configurations but there was a difference to the current `main` branch: In the `main` branch, the first frame was completely gray, the second completely blue. With the updates, the first three frames looked like this:


Frame 1:
Frame 1
Frame 2:
Frame 2
Frame 3:
Frame 3

In `main` the "Frame 2" was often the third frame, sometimes "Frame 1" was the third frame, followed by "Frame 2".

I believe that this is related to #140.

examples/platformer/main.rs
```diff
@@ -14,6 +14,7 @@ fn main() {
         .add_plugins(DefaultPlugins.set(ImagePlugin::default_nearest()))
         .add_plugin(LdtkPlugin)
         .add_plugin(RapierPhysicsPlugin::<NoUserData>::pixels_per_meter(100.0))
+        .configure_set(LdtkSystemSet::ProcessApi.before(PhysicsSet::SyncBackend))
```
Trouv, line 17: If this fixes the issue you observed we should probably leave a comment here explaining why it's necessary, and also mention it in the docs for `LdtkSystemSet::ProcessApi`

Trouv:
I see, it makes sense that we'd have to be explicit in the ordering of our systems vs rapiers now. I'm guessing with this explicit scheduling that the issue is truly gone, but it's hard to be sure. It's a bit of a shame to require users to do that - but I'm also okay with it for now. I think we could keep `0.6` focused on faithfully recreating the old schedule, and then later make more serious changes to the schedule with this issue and #140 in mind.

geieredgar:
Yeah, I believe this is something `bevy_ecs`'s ambiguity checker should be able to catch, but I think it is currently disabled by default because of the many false positives it reports. Improvements there might change that, so the user is at least warned about the race conditions. But this still leaves the question open which systems should run first, so I think it is important to document that somewhere. 

Another solution might be to add a `bevy_rapier` feature that adds the schedule constraint automatically in `LdtkPlugin::build`.

examples/platformer/main.rs
geieredgar: Added comments. Not sure if the `LdtkSystemSet::ProcessApi` comment is what you had in mind.

Trouv: I'd say explicitly mention `bevy_rapier` as an example since it's a common integration

Trouv:
> Another solution might be to add a `bevy_rapier` feature that adds the scheduling constraint automatically in `LdtkPlugin::build`.

Hmm, I think this could be cool if we had more integration with `bevy_rapier` as part of such a feature. For example, more `LdtkEntity` attribute macros for collisions, or something. My gut reaction to this is that users will see that feature listed and expect it to be more in depth than one scheduling constraint.

I think making users responsible for the scheduling constraint is fine for now but I definitely want to announce it in the release and have it documented somewhere like you've already done. I'm interested in reworking the scheduling a bit after this release anyway, so the requirement may disappear at that point

examples/platformer/main.rs
geieredgar: Done.

src/lib.rs
```diff
@@ -138,20 +138,11 @@ mod plugin {
...
-    pub enum LdtkStage {
-        /// Occurs immediately after [CoreStage::Update].
+    /// Base [SystemSet]s for systems added by the plugin.
+    #[derive(Copy, Clone, Eq, PartialEq, Debug, Hash, SystemSet)]
+    #[system_set(base)]
+    pub enum LdtkSystemSet {
+        /// Scheduled after [CoreSet::UpdateFlush].
         ///
         /// Used for systems that process components and resources provided by this plugin's API.
         /// In particular, this stage processes..
```
Trouv, line 148:
```suggestion
        /// In particular, this set processes..
```

src/lib.rs
```diff
@@ -138,20 +138,11 @@ mod plugin {
...
-    }
-
-    /// [StageLabel] for stages added by the plugin.
-    #[derive(Copy, Clone, Eq, PartialEq, Debug, Hash, StageLabel)]
-    pub enum LdtkStage {
-        /// Occurs immediately after [CoreStage::Update].
+    /// Base [SystemSet]s for systems added by the plugin.
+    #[derive(Copy, Clone, Eq, PartialEq, Debug, Hash, SystemSet)]
+    #[system_set(base)]
+    pub enum LdtkSystemSet {
```
Trouv, line 144: Now that it's more likely users will have to import this, could you export it in the prelude? After doing so we won't need to import it separately in the platformer example

src/lib.rs
```diff
@@ -161,10 +152,20 @@ mod plugin {
         /// - [components::Respawn]
         ///
         /// As a result, you can expect minimal frame delay when updating these in
-        /// [CoreStage::Update].
+        /// [CoreSet::Update].
```
Trouv, line 155: A similar `CoreStage` > `CoreSet` change needs to happen for the docs of `resources::LevelEvent::Transformed`

Trouv (commented):
A few more small things

src/lib.rs
geieredgar: Done.

src/lib.rs
geieredgar: Done.

Trouv (approved):
LGTM - I think I will merge this now just so people can depend on the main branch of the main repo rather than yours. We can fix the dependency once bevy_ecs_tilemap is updated.

Issue #176, opened 2023-04-07
Panic without `atlas` feature after v0.6.0 upgrade


johanhelsing:
I'm not really sure if this is intentional or not, but I get a panic after upgrading from v0.5.0 to v0.6.0 (on native).

```
2023-04-07T15:28:53.676302Z ERROR wgpu::backend::direct: Handling wgpu errors as fatal by default
thread 'Compute Task Pool (11)' panicked at 'wgpu error: Validation Error

Caused by:
    In Device::create_texture
      note: label = `texture_array`
    Dimension Z value 4096 exceeds the limit of 2048

', C:\Users\Johan\.cargo\registry\src\index.crates.io-6f17d22bba15001f\wgpu-0.15.1\src\backend\direct.rs:3024:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2023-04-07T15:28:53.677043Z  INFO cargo_space::ldtk_integration: Level event: Transformed("58e02ac0-5110-11ed-b070-cdccc74eaccd")
2023-04-07T15:28:53.677162Z  INFO cargo_space::ldtk_integration: Level event: Transformed("58e02ac0-5110-11ed-b070-cdccc74eaccd")
thread 'Compute Task Pool (11)' panicked at 'A system has panicked so the executor cannot continue.: RecvError', C:\Users\Johan\.cargo\git\checkouts\bevy-899673624796a25d\7b9124d\crates\bevy_ecs\src\schedule\executor\multi_threaded.rs:194:60
thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', C:\Users\Johan\.cargo\git\checkouts\bevy-899673624796a25d\7b9124d\crates\bevy_tasks\src\task_pool.rs:376:49
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: RecvError', C:\Users\Johan\.cargo\git\checkouts\bevy-899673624796a25d\7b9124d\crates\bevy_render\src\pipelined_rendering.rs:136:45
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', C:\Users\Johan\.cargo\git\checkouts\bevy-899673624796a25d\7b9124d\crates\bevy_tasks\src\task_pool.rs:376:49
error: process didn't exit successfully: `target\debug\cargo_space.exe --dev --join-private 123 --next 1` (exit code: 101)
```

The panics go away if I enable the `atlas` feature.

[...]

PR #183, opened 2023-04-19
fix!: create IntGrid white-image on asset load and minimize its size

### LINKED ISSUE #172: Cache white-image for entire asset
We had to create a white image for every level to support background colors and intgrid colors. The size of this image was previously kind of difficult to determine due to every level needing a minimum-sized image to encompass the entire level. The temporary solution was to just generate one for every level. This isn't ideal - it's a bit of wasted memory to store this many assets, especially when levels can get pretty big.

However, as of #171, we no longer need this white image for background layers, only intgrid colors. This means that the size of the image we need is just the maximum-sized intgrid layer in the asset. This is much easier to determine and create an image for early. I think we should create such an image as early as possible, maybe during asset loading (storing the handle on `LdtkAsset`).

Trouv:
Closes #172, and resolves an issue brought up by @Mattincho in #176 

# Summary
The white image generated for levels is now only used for intgrid coloring. This means it no longer needs to be the size of the level - it can just be the size of the maximum-sized intgrid layer. The maximum-sized intgrid tile is easy to determine during asset loading. This PR creates a small white-image for the maximum-sized intgrid tile once for the entire asset, rather than a large one once per level. This should improve performance, and also hopefully resolve issues with hitting hardware limits on larger levels.

BREAKING CHANGE: Most likely won't affect users - `LdtkAsset` has gained a `int_grid_image_handle` field, breaking any manual construction of it.

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -250,3 +250,113 @@ impl AssetLoader for LdtkLevelLoader {
         &["ldtkl"]
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::ldtk::{Definitions, LayerDefinition, Type};
+
+    use super::*;
+
+    #[test]
+    fn int_grid_image_is_white() {
+        let definitions = Definitions {
+            layers: vec![LayerDefinition {
+                purple_type: Type::IntGrid,
+                grid_size: 16,
+                ..default()
+            }],
+            ..default()
+        };
+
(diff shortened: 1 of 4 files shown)


Mattincho:
Tested the branch, and without enabling the atlas feature of bevy_ecs_tilemap, I can now load a level that doesn't work with main.

Issue #70, opened 2022-04-10
"relations" feature


Trouv:
With LDtk 1.0, field instances can now store a reference to another entity. Ever since, I've been thinking about how to properly allow users to transform these references into "relational" components that store an `Entity` object.

Here's the plan.
1. Create a new default feature called "relations" to make most of the following optional
2. Make `ldtk::FieldInstanceEntityReference` both `Hash` and `Component`, maybe rename it to just `EntityReference`
3. Give all LDtk entities spawned by the plugin an `EntityReference` component.
4. When spawning a level, create a `HashMap<EntityReference, Entity>` of all currently-spawned and soon-to-be-spawned LDtk entities, and pass it as a new argument into the `LdtkEntity` creation functions.
5. Create a new derivable `LdtkRelation` trait for constructing relational components.
6. Create an `#[entity]` field attribute for `LdtkRelation` for marking an `Entity` field, or maybe just make it smart enough to construct any `Entity` fields w/o an attribute.
7. Create a new `#[ldtk_relation("FIELD_IDENTIFIER")]` field attribute for the `LdtkEntity` derive that uses the `LdtkRelation` constructor to create the component from the entity reference field with the given field identifier.

Example:
```rust
#[derive(Component, LdtkRelation)]
struct MyRelation {
    #[entity]
    entity: Entity,
    other_info: i32,
}

#[derive(Bundle, LdtkEntity)]
struct MyBundle {
    #[ldtk_relation("my_reference_field_identifier")]
    relation: MyRelation,
    #[sprite_sheet_bundle]
    sprite_sheet_bundle: SpriteSheetBundle,
}
```

It might also be worth considering supporting deriving `LdtkRelation` on tuple structs, but I'm not exactly sure how to do that nicely.

Trouv:
As far as WHEN to implement all of this, it'd probably be good to aim for `0.3`, but that's a lot of work so I'd like to say at least get steps 1-4 in for `0.3` to allow users to manually create relational components in their custom `LdtkEntity` derives.

mystal:
Hi, I'm trying out `bevy_ecs_ldtk` this weekend for the first time and it's fantastic so far!

I was trying to figure out how to get `Entity`s from `FieldInstanceEntityReference`s, but it sounds like there isn't an easy way yet? I was starting to think I would tag all my loaded entities with a component that stored that information (similar to what you're proposing here).

[...]

PR #192, opened 2023-06-05
build a map of entity iid to EntityInstance during asset load

adtennant:
Simply to allow easier lookup for entities, e.g. from `EntityRef` fields.

It's a little brute-force and ideally the map should probably be a `HashMap<String, &EntityInstance>` to stop cloning the `EntityInstance` but `bevy` wasn't a fan of the lifetime - I also thought it was technically self-referencial but the compiler didn't seem upset by it.

It might be worth turning them into `Handle`s but that seems a little overkill as there could be a lot inside a map and then it gives you the lookup against `Assets<Handle<EntityInstance>>` in any system that wants to use it.

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -1,6 +1,6 @@
 //! Assets and AssetLoaders for loading ldtk files.
 
-use crate::{ldtk, resources::LevelSelection};
+use crate::{ldtk, resources::LevelSelection, EntityInstance};
 use bevy::{
     asset::{AssetLoader, AssetPath, LoadContext, LoadedAsset},
     prelude::*,
@@ -23,6 +23,8 @@ pub type TilesetMap = HashMap<i32, Handle<Image>>;
 /// Used in [LdtkAsset]. Key is the level iid.
 pub type LevelMap = HashMap<String, Handle<LdtkLevel>>;
 
+pub type EntityMap = HashMap<String, EntityInstance>;
+
 /// Main asset for loading ldtk files.
 ///
 /// Load your ldtk project with the asset server, then insert the handle into the
@@ -33,6 +35,7 @@ pub struct LdtkAsset {
     pub project: ldtk::LdtkJson,
     pub tileset_map: TilesetMap,
     pub level_map: LevelMap,
+    pub entity_map: EntityMap,
     /// Image used for rendering int grid colors.
     pub int_grid_image_handle: Option<Handle<Image>>,
 }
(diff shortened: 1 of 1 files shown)


Trouv:
Thanks for the contribution! This is actually extremely similar to something I've been thinking about doing but haven't gotten around to - and I would prefer to implement just one of these solutions. My thought was to basically create a resource that maps entity iids to to that entity's indices in the asset. I was also thinking, in conjunction with this, there could be an `EntityIid` component added to entities that is treated as the hash. And finally, maybe a custom `SystemParam` that provides a convenient api for this sort of lookup instead of users having to deal with `Assets<LdtkAsset>` and this resource themselves.

```rust
#[derive(Clone, Debug, PartialEq, Eq, Hash, Deref, Component)]
pub struct EntityIid(String);

#[derive(Copy, Clone, Debug, PartialEq, Eq)]
pub struct LdtkEntityIndices {
    world_index: usize,
    level_index: usize,
    layer_index: usize,
    entity_index: usize,
}

#[derive(Clone, Debug, PartialEq, Eq, Resource)]
pub struct LdtkEntityIndex {
    map: HashMap<Handle<LdtkAsset>, HashMap<EntityIid, LdtkEntityIndices>>,
}

#[derive(Clone, Debug, SystemParam)]
pub struct LdtkIndexer {
    assets: Res<Assets<LdtkAsset>>,
    index: Res<LdtkEntityIndex>
}

impl LdtkIndexer {
    // pub methods that provide convenient access *by reference* to levels, layers and entity instances for a given entity
    // some methods could be `_single()` and assume only one LdtkAsset exists since this is true for most users
}
```

The idea is that we won't be copying a decent sized type like `EntityInstance` that much, and that users can easily get access to the layer and level data an entity belongs to as well as the entity data itself. They already have easy access to these when spawning entities via the macro/registration system, but this makes it much easier for users that prefer the "blueprint" pattern of spawning entities to access this metadata, and for users to access this metadata well after spawning.

We could also maybe store an `Option<Entity>` on `LdtkEntityIndices` for making it easier for users to find already-spawned entities by entity iid. This will help considerably with #70.

What are your thoughts on this - do you think it would help you lookup entities in your use case?

adtennant:
The appraoch you've suggested makes sense to me and would enable me to do what I need to. I think as a starting point just having an `EntityIid` component on every `Entity` spawned by the plugin would make a massive difference.

Are you looking for contributions in this area? I quickly threw up these PRs yesterday to kick off some discussions as I couldn't see a `CONTRIBUTING` file and frankly I knew they were half-baked.

Trouv:
Sorry for the lack of `CONTRIBUTING`. I don't really have an official process for assigning tasks I suppose, I usually just make issues and assign them to myself when I begin working on them. Unfortunately, I don't have an issue filed for what I've just described here.
That being said, I probably won't be doing feature work on this plugin for a couple weeks or so so feel free to work on it if you want to! No pressure of course, but contributions are always appreciated. `EntityIid` does sound like a good first step, I agree.

adtennant:
Having quickly implemented adding `EntityIid` to everything, I am wondering if instead having `EntityInstance` not get removed when the type is mapped might be more elegant. It would open up the possibility of using `Added<EntityInstance>` in combination with the `LdtkEntity` trait as currently they are mutally exclusive.

That said, with the lookup approach then `Added<EntityIid>` opens up as a possibility, which means not needing to copy `EntityInstance` at all. On second thoughts - that's probably the better way of going.

PR #193, opened 2023-06-05
early exit in apply_level_selection if nothing would cause the LevelSet to change

adtennant:
I'm sure there are some pitfalls here, but I've been trying to extend the `LevelSelection` behaviour, and rather than completely replace it (because I want neighbours, and why rewrite it?) I've been trying to extend it by shoving a system in the middle.

```
apply_level_selection
    .in_base_set(bevy_ecs_ldtk::LdtkSystemSet::ProcessApi)
    .after(bevy_ecs_ldtk::systems::apply_level_selection)
    .before(bevy_ecs_ldtk::systems::apply_level_set),
```

In doing so I noticed the `bevy_ecs_ldtk::systems::apply_level_selection` system runs every single frame, even when nothing should be being updated. This meant that my system also had to run every frame - in order to stop the changed being overridden - except my calculations were considerably more expensive so I really didn't want that to happen.

As with my other PR, it's a little brute-force but hopefully a starting point for discussion.

Sidenote: should `LevelSelection` be a component rather than a `Resource`? Similar to `LevelSet`.

### DIFF
--- a/src/systems.rs
+++ b/src/systems.rs
@@ -73,6 +73,11 @@ pub fn apply_level_selection(
     #[cfg(feature = "render")] mut clear_color: ResMut<ClearColor>,
 ) {
     if let Some(level_selection) = level_selection {
+        if !level_selection.is_changed() && !ldtk_settings.is_changed() && !ldtk_assets.is_changed()
+        {
+            return;
+        }
+
         for (ldtk_handle, mut level_set) in level_set_query.iter_mut() {
             if let Some(ldtk_asset) = ldtk_assets.get(ldtk_handle) {
                 if let Some(level) = ldtk_asset.get_level(&level_selection) {


src/systems.rs
```diff
@@ -73,6 +73,11 @@ pub fn apply_level_selection(
     #[cfg(feature = "render")] mut clear_color: ResMut<ClearColor>,
 ) {
     if let Some(level_selection) = level_selection {
+        if !level_selection.is_changed() && !ldtk_settings.is_changed() && !ldtk_assets.is_changed()
```
Trouv, line 76: I think this is the sort of thing better handled by bevy run criteria for a potential small perf benefit - it could even be pretty generic like `.run_if(resource_exists::<LevelSelection>).and_then(resource_changed::<LevelSelection>)...` I'm not sure if I care about change detection for the settings though, those are definitely conceptually more of an immutable thing.

Trouv (changes requested):
This *seems* like it should be fine, but there were an awful lot of bugs when I was originally developing this around this sort of assumption. I'll test this a good amount before merging.

src/systems.rs
adtennant: I think on reflection this is better handled by making `LevelSelection` a component that is matched to the `Handle<LdtkAsset>` and available in `LdtkWorldBundle`, the change handling is a little less dicey then. It also allows more granular control when rendering multiple worlds. The same holds for the settings.

Regarding the settings specifically, IMO whilst they may be conceptually immutable that isn't something that's enforced by bevy and therefore should be taken into account.

Same question re: contributions that I asked here:
https://github.com/Trouv/bevy_ecs_ldtk/pull/192#issuecomment-1578776605

Trouv: > I think on reflection this is better handled by making `LevelSelection` a component that is matched to the `Handle<LdtkAsset>` and available in `LdtkWorldBundle`, the change handling is a little less dicey then. It also allows more granular control when rendering multiple worlds. The same holds true for the settings.

hmm.. I like having `LevelSelection` as a resource since it's convenient for most users that only have one world to worry about. I think I would be okay with it being both though. So the workflow would be
- an optional `LevelSelection` resource and a system that updates all `LevelSelection` components in the world to sync with the resource if it exists. It could also add `LevelSelection` components to `Handle<LdtkAsset>` entities that don't have one.
- `LevelSelection` component and a system that updates that entity's corresponding `LevelSet` component from it. Exclude it from `LdtkWorldBundle`.
- a system that spawns levels as children of `LevelSet` entities, and responds to updates in an idempotent way (this is the current behavior)

So users that have multiple worlds can still use `LevelSelection` conveniences by using it as a component, but most users can keep using it as a resource. This change probably deserves its own PR.

Thoughts?

adtennant: That all makes perfect sense to me. How do you want this approached? An issue with an RFC-like summary? Or are you happy for it to just come in as a fresh PR with a summary in the description.

PR #194, opened 2023-06-08
feat: add `EntityIid` component which is added to all entities by default

adtennant:
This is a first step on the road to having an easy way of looking up entity information from `LdtkAsset`.

**Questions**
* ~~Should it be possible to disable this using `LdtkSettings`? I'm not sure it's worth worrying about as the memory usage should be minimal as the IID are always GUIDs. Perhaps a feature?~~
* ~~Should this be a newtype? It makes sense to me and I can't see any reason for it to ever gain fields, but the other components within the crate all have named fields.~~
* ~~Should the `EntityIid` also carry its `Handle<LdtkAsset>`? I don’t think so personally as `LevelSelection` doesn’t have this concept.~~

### DIFF
--- a/src/components.rs
+++ b/src/components.rs
@@ -5,6 +5,7 @@ use crate::ldtk::{LayerInstance, Type};
 use bevy::prelude::*;
 
 use std::{
+    borrow::Cow,
     collections::HashSet,
     ops::{Add, AddAssign, Mul, MulAssign, Sub, SubAssign},
 };
@@ -18,6 +19,60 @@ use crate::{
 };
 
 use bevy_ecs_tilemap::tiles::{TileBundle, TilePos};
+
+/// [Component] added to all [LdtkEntity]s by default.
+///
+/// The `iid` stored in this component can be used to uniquely identify LDtk entities within an [LdtkAsset].
+#[derive(Clone, Debug, Default, Deref, Hash, Eq, PartialEq, Component, FromReflect, Reflect)]
+#[reflect(Component, Default, Debug)]
+pub struct EntityIid(Cow<'static, str>);
+
+impl EntityIid {
+    pub fn new(iid: impl Into<Cow<'static, str>>) -> Self {
+        let iid = iid.into();
+        EntityIid(iid)
+    }
+
+    #[inline(always)]
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+}
+
+impl From<&str> for EntityIid {
+    #[inline(always)]
+    fn from(value: &str) -> Self {
+        EntityIid::new(value.to_owned())
+    }
+}
+
+impl From<String> for EntityIid {
+    #[inline(always)]
+    fn from(value: String) -> Self {
+        EntityIid::new(value)
+    }
+}
+
+impl AsRef<str> for EntityIid {
+    #[inline(always)]
+    fn as_ref(&self) -> &str {
+        &self.0
+    }
+}
+
+impl From<&EntityIid> for String {
+    #[inline(always)]
+    fn from(value: &EntityIid) -> String {
+        value.as_str().to_owned()
+    }
+}
+
+impl From<EntityIid> for String {
+    #[inline(always)]
+    fn from(value: EntityIid) -> String {
+        value.0.into_owned()
+    }
+}
 
 /// [Component] added to any `IntGrid` tile by default.
 ///
--- a/src/components/entity_iid.rs
+++ b/src/components/entity_iid.rs
@@ -10,11 +10,13 @@ use std::borrow::Cow;
 pub struct EntityIid(Cow<'static, str>);
 
 impl EntityIid {
+    /// Construct an [`EntityIid`].
     pub fn new(iid: impl Into<Cow<'static, str>>) -> Self {
         let iid = iid.into();
         EntityIid(iid)
     }
 
+    /// Returns the internal entity iid as a `&str`.
     #[inline(always)]
     pub fn as_str(&self) -> &str {
         &self.0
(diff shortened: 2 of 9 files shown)


src/components.rs
```diff
@@ -19,6 +20,60 @@ use crate::{
 
 use bevy_ecs_tilemap::tiles::{TileBundle, TilePos};
 
+/// [Component] added to all [LdtkEntity]s by default.
+///
+/// The `iid` stored in this component can be used to uniquely identify LDtk entities within an [LdtkAsset].
+#[derive(Clone, Debug, Default, Deref, DerefMut, Hash, Eq, PartialEq, Component, Reflect)]
+#[reflect(Component)]
+pub struct EntityIid(Cow<'static, str>);
```
Trouv, line 28: I'm not that experienced with the `Cow` pointer, but since the internal value here isn't pub and we only provide immutable access to it, I'm not sure what the benefit is of using it? I'm not saying there isn't one, just curious what the thought process was here

src/components.rs
```diff
@@ -19,6 +20,60 @@ use crate::{
...
+        value.as_str().to_owned()
+    }
+}
+
+impl From<EntityIid> for String {
+    #[inline(always)]
+    fn from(value: EntityIid) -> String {
+        value.0.into_owned()
+    }
+}
```
Trouv, line 75: I know this violates the current organization of the repository, but I'd like to reorganize things soon(tm) into smaller modules. Do you think you could put all of this `EntityIid` stuff into its own module?

Trouv (changes requested):

Trouv:
> * Should it be possible to disable this using `LdtkSettings`? I'm not sure it's worth worrying about as the memory usage should be minimal as the IID are always GUIDs. Perhaps a feature?

I think the `LdtkIndexer` stuff I mentioned makes sense to go behind a feature, but this `EntityIid` is small enough and generally-useful-enough that I think it's fine to make it non-optional.

> * Should this be a newtype? It makes sense to me and I can't see any reason for it to ever gain fields, but the other components within the crate all have named fields.

That's fine - I was pretty allergic to newtypes when I first wrote this plugin but no longer.

> * Should the `EntityIid` also carry its `Handle<LdtkAsset>`? I don’t think so personally as `LevelSelection` doesn’t have this concept.

I agree, it doesn't need it.

src/components.rs
adtennant: Mostly I was following the lead of [bevy](https://docs.rs/bevy/latest/bevy/index.html)::[core](https://docs.rs/bevy/latest/bevy/core/index.html)::[Name](https://docs.rs/bevy/latest/bevy/core/struct.Name.html#) as I see these being very similar (https://github.com/bevyengine/bevy/blob/main/crates/bevy_core/src/name.rs). 

There probably isn't a ton of benefit to it although it does make all the conversions and `new` trivial to implement.

adtennant:
@Trouv I've refactored `components.rs` into a folder, as I assume given time you'll want to break it down fully. It seemed neater than having `entity_iid.rs` floating around on its own.

src/components/entity_iid.rs
```diff
@@ -0,0 +1,57 @@
+use bevy::prelude::*;
+
+use std::borrow::Cow;
+
+/// [Component] added to all [LdtkEntity]s by default.
+///
+/// The `iid` stored in this component can be used to uniquely identify LDtk entities within an [LdtkAsset].
+#[derive(Clone, Debug, Default, Deref, DerefMut, Hash, Eq, PartialEq, Component, Reflect)]
```
Trouv, line 8: I prefer to not provide mutable access to the internal type here 
```suggestion
#[derive(Clone, Debug, Default, Deref, Hash, Eq, PartialEq, Component, Reflect)]
```

src/components.rs
Trouv: I see. Well if it's good enough for bevy, it's good enough for me

Trouv (changes requested):

src/components/entity_iid.rs
```diff
@@ -0,0 +1,57 @@
...
+pub struct EntityIid(Cow<'static, str>);
+
+impl EntityIid {
+    pub fn new(iid: impl Into<Cow<'static, str>>) -> Self {
+        let iid = iid.into();
+        EntityIid(iid)
+    }
+
+    #[inline(always)]
+    pub fn as_str(&self) -> &str {
```
Trouv, line 19:
```suggestion
    /// Returns the internal entity iid as a `&str`.
    #[inline(always)]
    pub fn as_str(&self) -> &str {
```

src/components/entity_iid.rs
```diff
@@ -0,0 +1,57 @@
...
+
+/// [Component] added to all [LdtkEntity]s by default.
+///
+/// The `iid` stored in this component can be used to uniquely identify LDtk entities within an [LdtkAsset].
+#[derive(Clone, Debug, Default, Deref, Hash, Eq, PartialEq, Component, FromReflect, Reflect)]
+#[reflect(Component, Default, Debug)]
+pub struct EntityIid(Cow<'static, str>);
+
+impl EntityIid {
+    pub fn new(iid: impl Into<Cow<'static, str>>) -> Self {
```
Trouv, line 13:
```suggestion
    /// Construct an [`EntityIid`].
    pub fn new(iid: impl Into<Cow<'static, str>>) -> Self {
```

Trouv (changes requested):
Sorry to be pedantic - I'd like the public items to have doc-strings

Trouv (changes requested):
The `field_instances` example should probably be updated to take advantage of this in `mother.rs`

adtennant:
No worries about being pedantic, that is the point of doing code reviews 👍

Trouv (approved):
LGTM, thanks again

PR #195, opened 2023-06-08
feat: add `get_entity_instance` to `LdtkAsset`

adtennant:
This builds on the work in https://github.com/Trouv/bevy_ecs_ldtk/pull/194. I've split into two PRs in order to keep discussion focused, and because I think https://github.com/Trouv/bevy_ecs_ldtk/pull/194 in isolation still gives some benefits if this PR takes a little longer to get over the line. Unfortunately I can only set the base branch to one which exists in this repo (rather than my fork) so the diff includes all the other changes for now.

I went down this route rather than creating a `SystemParam` as discussed in https://github.com/Trouv/bevy_ecs_ldtk/pull/192 as it felt more natural for it to live alongside `get_level` and it massively simplifies the implementation.

I'm going to give some thought to how this can be extended to enable https://github.com/Trouv/bevy_ecs_ldtk/issues/70 as that is closely related to my actual use-case.

**Questions**
* Should the `traitless` example be updated to use this? Full trait less will still work the same, but with this you can mix both approaches. It removes a potential footgun.

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -6,9 +6,9 @@ use bevy::{
     prelude::*,
     reflect::TypeUuid,
     render::render_resource::{Extent3d, TextureDimension, TextureFormat},
-    utils::BoxedFuture,
+    utils::{BoxedFuture, HashMap},
 };
-use std::{collections::HashMap, path::Path};
+use std::path::Path;
 
 #[allow(unused_imports)]
 use crate::components::LdtkWorldBundle;
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -1,6 +1,10 @@
 //! Assets and AssetLoaders for loading ldtk files.
 
-use crate::{ldtk, resources::LevelSelection};
+use crate::{
+    ldtk::{self, Level},
+    resources::LevelSelection,
+    EntityIid, EntityInstance,
+};
 use bevy::{
     asset::{AssetLoader, AssetPath, LoadContext, LoadedAsset},
     prelude::*,
@@ -23,6 +27,18 @@ pub type TilesetMap = HashMap<i32, Handle<Image>>;
 /// Used in [LdtkAsset]. Key is the level iid.
 pub type LevelMap = HashMap<String, Handle<LdtkLevel>>;
 
+/// Used in [LdtkAsset]. Enables faster lookups in [LdtkAsset::get_entity_instance()].
+#[derive(Clone, Debug, Default, Eq, Hash, PartialEq, Component, Reflect)]
+pub struct EntityIndexes {
+    world_index: Option<usize>,
+    level_index: usize,
+    layer_index: usize,
+    entity_index: usize,
+}
+
+/// Used in [LdtkAsset]. Enables faster lookups in [LdtkAsset::get_entity_instance()].
+pub type EntityIndex = HashMap<EntityIid, EntityIndexes>;
+
 /// Main asset for loading ldtk files.
 ///
 /// Load your ldtk project with the asset server, then insert the handle into the
@@ -33,6 +49,7 @@ pub struct LdtkAsset {
     pub project: ldtk::LdtkJson,
     pub tileset_map: TilesetMap,
     pub level_map: LevelMap,
+    pub entity_index: EntityIndex,
     /// Image used for rendering int grid colors.
     pub int_grid_image_handle: Option<Handle<Image>>,
 }
@@ -107,6 +124,21 @@ impl LdtkAsset {
             .find(|(i, l)| level_selection.is_match(i, l))
             .map(|(_, l)| l)
     }
+
+    /// Find a particular entity instance using an [EntityIid].
+    pub fn get_entity_instance(&self, entity_iid: &EntityIid) -> Option<&EntityInstance> {
+        let indexes = self.entity_index.get(entity_iid)?;
+
+        let levels = if let Some(world_index) = indexes.world_index {
+            &self.project.worlds.get(world_index)?.levels
+        } else {
+            &self.project.levels
+        };
+
+        let level = levels.get(indexes.level_index)?;
+        let layer_instance = level.layer_instances.as_ref()?.get(indexes.layer_index)?;
+        layer_instance.entity_instances.get(indexes.entity_index)
+    }
 }
 
 #[derive(Copy, Clone, Debug, Default)]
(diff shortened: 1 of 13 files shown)


Trouv:
> I went down this route rather than creating a SystemParam as discussed in https://github.com/Trouv/bevy_ecs_ldtk/pull/192 as it felt more natural for it to live alongside get_level and it massively simplifies the implementation.

Hmm I see what you mean. 

> I'm going to give some thought to how this can be extended to enable https://github.com/Trouv/bevy_ecs_ldtk/issues/70 as that is closely related to my actual use-case.

Yeah, editing the asset data is a no-no, and the spawned `Entity`s for an ldtk world naturally changes as the levels load/unload. I think, going with the approach you've laid out here, having a separate feature where we map `EntityIid`s to `Entity`s in a resource makes sense. `Option<Entity>` didn't really feel like it totally fit in `EntityIndices` when I originally suggested it anyway.

> * Should the `traitless` example be updated to use this? Full trait less will still work the same, but with this you can mix both approaches. It removes a potential footgun.

I think using this in an example is a good idea, but I'm fine with leaving that for a future PR. I'm not sure if `traitless` is the right one to update - its intention is to basically be a `--no-default-features` version of the `basic` example.

adtennant:
@Trouv I’ll rebase this in top of `main` tomorrow now that https://github.com/Trouv/bevy_ecs_ldtk/pull/194 is merged.

Do you have any feedback specific to these changes that needs addressing? Or would you rather make until there is a clean diff and discuss further?

src/assets.rs
```diff
@@ -107,6 +124,21 @@ impl LdtkAsset {
...
+        let levels = if let Some(world_index) = indexes.world_index {
+            &self.project.worlds.get(world_index)?.levels
+        } else {
+            &self.project.levels
+        };
+
+        let level = levels.get(indexes.level_index)?;
+        let layer_instance = level.layer_instances.as_ref()?.get(indexes.layer_index)?;
+        layer_instance.entity_instances.get(indexes.entity_index)
+    }
```
Trouv, line 141: I would like there to be separate methods for getting a reference to the layer instance and the level as well. It's okay to keep this PR small without them - but if they could depend on each other somehow then it might make sense to implement them all here and it would probably be an ideal implementation.

However, I think it's not quite that simple because we want these methods to accept an `&EntityIid` not an `&EntityIndexes`. So - if we had `get_entity_instance` use `get_layer_for_entity` internally, which in turn uses `get_level_for_entity` internally, we'd actually be looking up the same value in `EntityIndex` 3 times. I guess that's not that big of a deal though because the lookup is constant.

src/assets.rs
```diff
@@ -107,6 +124,21 @@ impl LdtkAsset {
             .find(|(i, l)| level_selection.is_match(i, l))
             .map(|(_, l)| l)
     }
+
+    /// Find a particular entity instance using an [EntityIid].
```
Trouv, line 128: It's kind of obvious, but we should explain what a `None` (or as suggested in my next comment, `Err`) result means.

src/assets.rs
```diff
@@ -107,6 +124,21 @@ impl LdtkAsset {
...
+    pub fn get_entity_instance(&self, entity_iid: &EntityIid) -> Option<&EntityInstance> {
```
Trouv, line 129: I wonder if returning `Result` w/ a custom error would be more clear. As opposed to a lookup in a `HashMap` where there are plenty of use cases that users would perform some logic in a `None` result, I think here users are more likely to be expecting this entity to exist.

A simple thiserror `NoSuchEntity` tuple-struct would suffice in the `Err` variant:
```rust
#[derive(Debug, Error)]
#[error("provided entity iid doesn't exist in the ldtk asset's index: {0}")]
struct NoSuchEntity(EntityIid);
```

I believe this is the only error we should be worried about. We should be able to `.expect()` the indexing of levels, layers, and entities if an `EntityIndexes` exists. The logic for creating that map is all private and should be correct

src/assets.rs
```diff
@@ -33,6 +49,7 @@ pub struct LdtkAsset {
     pub project: ldtk::LdtkJson,
     pub tileset_map: TilesetMap,
     pub level_map: LevelMap,
+    pub entity_index: EntityIndex,
```
Trouv, line 52: As mentioned in #192, I think it makes sense to put this field (and the types it depends on) behind a default `"index"` feature.

src/assets.rs
```diff
@@ -176,10 +208,53 @@ impl AssetLoader for LdtkLoader {
...
             });
 
+            fn build_entity_index_for_levels(
+                world_index: Option<usize>,
+                levels: &[Level],
+            ) -> EntityIndex {
+                let mut entity_index = HashMap::new();
+
+                for (level_index, level) in levels.iter().enumerate() {
+                    if let Some(layer_instances) = level.layer_instances.as_ref() {
```
Trouv, line 218: Ah, - here's a bit of a snag in the implementation. These layers won't exist when LDtk is set to use external level files. They'll only exist on the `LdtkLevel` asset in that case. So, the entity index will be completely empty for users using external levels.

For context - if the user is not using external level files - those `LdtkLevel` assets *will* still exist and they'll be set as "labeled" assets of the `LdtkAsset`. If the user is using external level files - those `LdtkLevel` assets will be loaded separately as "dependent" assets of the `LdtkAsset`. Either way - the `Handle<LdtkLevel>` gets stored on the `LdtkAsset::level_map`.

I think we can only access the level data from this function in the "internal" case, unfortunately. In the external case, I believe the level data is loaded completely asynchronously to this function - and we can't await on it.

I think to have this feature work in both cases - we'll need access to both `LdtkAsset` data *and* `LdtkLevel` data - which I believe means we'll need to perform this logic during systems that access `Assets<LdtkAsset>` and `Assets<LdtkLevel>`, which is bringing me back to the `SystemParam` idea. Though, if you have another idea, or I'm missing something about bevy's asset system that allows us to do it here, I'm happy to hear it.

Trouv (changes requested):
This almost works, but unfortunately I forgot about external level files. I recommend reading my last comment first as addressing that probably changes things more fundamentally.

src/assets.rs
adtennant: Ah, yes. 

Is another potential solution to address this in documentation in the same way `get_level` works, or would you rather jump all the way to the most robust solution?

We'd need to provide some way of looking up within the `LdtkLevel` as well if we went down the documentation route.

The above comes from a slightly selfish place in that whilst the more robust (likely `SystemParam`) solution is probably better for the plugin longer term, I only need the more niave solution for my needs. I'm still happy to keep this moving, but if I can unblock my main project sooner that's obviously better for me - although there's nothing stopping me using a private fork for now to solve that.

adtennant:
Bumping this back to draft whilst we resolve what this should look like, @Trouv would you prefer me to create an issue to discuss this?

src/assets.rs
Trouv: Hmm my thought here is - is the naive solution a partial implementation of the full solution? Like is it something we can build on top of for future releases or is it something we'd like to replace in future releases? The way I imagine things working, I think it would be the latter, which makes me hesitant to merge.

Now what might be a good partial implementation would be to have a map of entityiids to layer/entity indexes for a specific level generated as a new field on an `LdtkLevel`. So, the same strategy you've implemented here but smaller in scope. I think then generating a resource mapping `EntityIid`s to `Handle<LdtkLevel>`s could build naturally on top of that, and a `SystemParam` providing a convenient API could build on top of that. I'm not sure if that first step would be very useful to you?

src/assets.rs
Trouv: I see - I'm okay with that, this was a light suggestion anyway. Though the 2nd part of this suggestion makes me wonder - is it useful to `.expect()` something expected where a simple `?` is compatible? Probably slightly for catching bugs but not a big deal

Trouv:
> Bumping this back to draft whilst we resolve what this should look like, @Trouv would you prefer me to create an issue to discuss this?

I've been reworking the asset types and this should be a little bit easier now. I finally created an issue for this and am closing this PR in favor of it: #251. Thanks again, @adtennant

PR #201, opened 2023-07-09
add reflect too LdtkLevel and all parts

Hellzbellz123:
closed earlier commit and reopened with my main branch and commits
derives Reflect and FromReflect on the necessary parts for LdtkLevel and i added too World too but its not used
mostly just used too visually debug things in egui_inspector, but can be useful elswhere.

i have reflect_ignore on the real_editor_values Vec<serde_json::Value> because its external

i need a git crash course,
i tried too rebase from main but it would then ask me too pull down changes from my fork, i just rebased my branches main kinda on accident but its fixed now

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -27,7 +27,7 @@ pub type LevelMap = HashMap<String, Handle<LdtkLevel>>;
 ///
 /// Load your ldtk project with the asset server, then insert the handle into the
 /// [LdtkWorldBundle].
-#[derive(Clone, TypeUuid, Reflect, FromReflect)]
+#[derive(Clone, TypeUuid)]
 #[uuid = "ecfb87b7-9cd9-4970-8482-f2f68b770d31"]
 pub struct LdtkAsset {
     pub project: ldtk::LdtkJson,
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -203,7 +203,7 @@ impl AssetLoader for LdtkLoader {
 /// The label is just the level's identifier.
 ///
 /// Loaded as a dependency to the [LdtkAsset] when loading an ldtk file with external levels.
-#[derive(TypeUuid)]
+#[derive(TypeUuid, Reflect, FromReflect)]
 #[uuid = "5448469b-2134-44f5-a86c-a7b829f70a0c"]
 pub struct LdtkLevel {
     pub level: ldtk::Level,
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -27,7 +27,7 @@ pub type LevelMap = HashMap<String, Handle<LdtkLevel>>;
 ///
 /// Load your ldtk project with the asset server, then insert the handle into the
 /// [LdtkWorldBundle].
-#[derive(Clone, TypeUuid)]
+#[derive(Clone, TypeUuid, Reflect, FromReflect)]
 #[uuid = "ecfb87b7-9cd9-4970-8482-f2f68b770d31"]
 pub struct LdtkAsset {
     pub project: ldtk::LdtkJson,
--- a/src/ldtk/mod.rs
+++ b/src/ldtk/mod.rs
@@ -25,7 +25,7 @@
 //! 13. All urls in docs have been changed to hyperlinks with `<>`
 //! 14. `Reflect` has been derived for [Type].
 
-use bevy::prelude::{Color, IVec2, Vec2};
+use bevy::{prelude::{Color, IVec2, Vec2}, reflect::{FromReflect, Reflect}};
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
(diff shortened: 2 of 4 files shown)


Trouv:
>i need a git crash course,
i tried too rebase from main but it would then ask me too pull down changes from my fork, i just rebased my branches main kinda on accident but its fixed now

I see! No worries - git is a complicated thing, and what you've done here accomplishes what I was asking for!

src/ldtk/mod.rs
```diff
@@ -969,7 +974,7 @@ pub struct ForcedRefs {
 /// When loading levels, you can flesh out LDtk entities in your own system by querying for
 /// `Added<EntityInstance>`.
 /// Or, you can hook into the entity's spawning process using [LdtkEntity].
-#[derive(PartialEq, Debug, Default, Clone, Serialize, Deserialize, bevy::prelude::Component)]
+#[derive(PartialEq, Debug, Default, Clone, Serialize, Deserialize, bevy::prelude::Component, Reflect, FromReflect)]
```
Trouv, line 977: Could you run `cargo fmt` to please the CI?

src/assets.rs
```diff
@@ -203,7 +203,7 @@ impl AssetLoader for LdtkLoader {
 /// The label is just the level's identifier.
 ///
 /// Loaded as a dependency to the [LdtkAsset] when loading an ldtk file with external levels.
-#[derive(TypeUuid)]
+#[derive(TypeUuid, Reflect, FromReflect)]
 #[uuid = "5448469b-2134-44f5-a86c-a7b829f70a0c"]
 pub struct LdtkLevel {
```
Trouv, line 208: Curious what you think about adding this to `LdtkAsset` too?

Trouv (changes requested):

Trouv:
BTW - it sounds like your forks main branch pulled this forks 0.8 branch is that right? So it makes sense that `git rebase -i main` wouldn't really work since you're developing on your fork's main branch, and even if you weren't developing there, your main branch still had those changes and such a rebase wouldn't give you the option to exclude them. I have a couple tips for this..

First, you can undo a pull on your main branch. A branch is essentially just a name pointing to some commit in the repository, and you can change what commit a branch is pointing to via a "reset". This is kind of a dangerous operation, but you could run something like `git reset HEAD^ --hard` to point the branch you're using to the previous commit (`HEAD^` is a short name for the previous commit), essentially deleting the most recent commit. Obviously you can lose work this way, so be careful using it.

Also, to help me organize my forks, I like to have a couple "remotes" set up. One remote "origin" pointing to my fork (which you pretty much get by default when you clone your fork), and then one remote "upstream" pointing to the main repository. You can add the "upstream" remote pretty easily using whichever link you'd use to clone this repository:
```
git remote add upstream git@github.com:Trouv/bevy_ecs_ldtk.git
```
And from there you can interact with upstream branches, rather than just your local ones or your fork's branches. So, you could probably have accomplished that rebase command I suggested without any resetting by changing it to `git rebase -i upstream/main`.

Hope this is good for future reference, sorry if some of this is stuff you already know.

src/assets.rs
Hellzbellz123: didnt have it on the asset because in my project in iterating over the Assets<LdtkLevel> but that is also something that should be done

Trouv:
So, the `field_instances` example actually provides a `bevy_inspector_egui` window, and I'm not seeing the asset in there still. I think maybe you need to register some of these types to the app as well via [`App::register_type`](https://docs.rs/bevy/latest/bevy/app/struct.App.html#method.register_type).

I'm not completely sure how it works, maybe you can just register `LdtkAsset` and `LdtkLevel` and that's all you need? It's possible some of the dependent types would need to be registered too, but I'd like to minimize that if possible.

Hellzbellz123:
My workstation died, currently setting up a new environment

Pretty sure everything on ldtklevel and level will need to be registered. 
I don't think .register_type<T>() works recursively yet
it might be useful too reflect ignore more types depending on the data and its usefulness but I'm pretty sure we got everything that should be ignored already

Trouv:
Turns out the method we need is [`App::register_asset_reflect`](https://docs.rs/bevy_asset/0.10.1/bevy_asset/trait.AddAsset.html#tymethod.register_asset_reflect), and we only need to use it on the top-level.

Interestingly, when I go to expand the `LdtkAsset` in that example, apparently the level despawns as I'm running into [this `.expect()`](https://github.com/Trouv/bevy_ecs_ldtk/blob/main/examples/field_instances/level_title.rs#L22) with an error `NoEntities`. It seems that expanding triggers an `AssetModified` event, and we handle these events by respawning the world, which probably isn't the correct behavior.

I think this needs to be addressed somehow before merging. We could change the `AssetModified` event handling, but that sounds like it will be difficult and could easily introduce bugs. Instead, I'll take back my previous suggestion of implementing `Reflect` on `LdtkAsset`, and ask to leave it on `LdtkLevel` only, since expanding `LdtkLevel` doesn't have this problem. We can take more drastic measures in future PRs if inspecting the `LdtkAsset` is something people want to do.

Trouv (approved):
Went ahead and implemented those changes since I want to deal with some inevitable merge conflicts this weekend. Thanks again for the contribution!

PR #204, opened 2023-07-23
Update to Bevy 0.11

PhaestusFox:
I don't know if this works 100% as I don't use the plugin myself and am updating for a Patreon, I just sort of guessed the ordering for what used to be base sets and now need a hard scheduler. let me know if something is in the wrong order and ill fix it


src/plugin.rs
```diff
@@ -40,18 +39,18 @@ impl Plugin for LdtkPlugin {
     fn build(&self, mut app: &mut App) {
         // Check if we have added the TileMap plugin
         if !app.is_plugin_added::<bevy_ecs_tilemap::TilemapPlugin>() {
-            app = app.add_plugin(bevy_ecs_tilemap::TilemapPlugin);
+            app = app.add_plugins(bevy_ecs_tilemap::TilemapPlugin);
         }
 
         app.configure_set(
+            PostUpdate,
```
Trouv, line 46: Unfortunately, it's actually pretty important for there to be an `apply_deferred` between `ProcessApi` and the ["removal helpers"](https://github.com/StarArawn/bevy_ecs_tilemap/blob/main/src/render/mod.rs#L133) in `bevy_ecs_tilemap`. Since these systems are private, we can't eplictly insert something `.before(..)` them, so `ProcessApi` just needs to go before `PostUpdate` entirely. I also put it after `Update` so that users using the ecs "api" in `Update` would notice no frame delay. 

This isn't ideal and now that stageless is here we can probably design something better - but for the purposes of this PR, I think it's fine to try and keep the behavior the same as it was before. So, I think there needs to be a custom schedule inserted after `Update` using the [`MainScheduleOrder`](https://docs.rs/bevy/latest/bevy/app/struct.MainScheduleOrder.html#method.insert_after) resource.

Trouv (changes requested):

Trouv:
@PhaestusFox Sorry I neglected to thank you for this contribution, it's very helpful. Do you mind if I wrap this up for you? I'd like to release 0.8 this weekend but since you said this was patreon-supported work I wouldn't want to get in the way of a video you're making or something.

Btw, that's really cool to me that someone requested via patreon that you update some repository you otherwise don't have any involvement with. I never even considered that a possible avenue of patreon work. I love it!

yornaath:
Eagerly anticipating this so I cant start my project :D

Trouv (approved):
Thanks again. I went ahead and implemented my own suggestions here, and ran into/fixed a couple other issues.  Users reading this should be aware that you'll likely run into some `bevy_ecs_tilemap` 0.11 bugs if you use this before 0.11.1 releases. These are fixed on `bevy_ecs_tilemap` main, which you can patch in using this:

```toml
[patch.crates-io]
bevy_ecs_tilemap = { git = "http://github.com/StarArawn/bevy_ecs_tilemap", version = "0.11", branch = "main" }
```

PR #206, opened 2023-07-29
refactor!: restructure assets module

Trouv:
The code quality of `assets.rs` was lacking. It was a bit too large, too public, under-documented, and overly-responsible. This PR makes a variety of changes mostly aimed at improving the assets module without changing core functionality.

This will likely be merged after the `0.8` release, and released in `0.9` instead (between bevy upgrades). This is so users can deal with the breaking changes at their own pace rather than having to deal with them when upgrading their projects to bevy `0.11`. Though, the breaking changes here shouldn't affect users much under normal usage.

# Non-breaking changes
- New `LdtkAssetPlugin` moves the registration of the asset types out of `plugin.rs`
- `LdtkProject` and `LdtkLevel` are given their own submodules of `assets`.
- Some extension methods added to `ldtk` types are now given their own "`impl`" submodules in `ldtk`.

# Breaking changes
BREAKING-CHANGE: `LdtkAsset` renamed to `LdtkProject`
BREAKING-CHANGE: `LdtkProject::project` and `LdtkLevel::level` fields have both been renamed to `data`
BREAKING-CHANGE: All fields of `LdtkProject` and `LdtkLevel` are now privatized, and have immutable getter methods
BREAKING-CHANGE: `LevelMap` and `TilesetMap` type aliases have been removed
BREAKING-CHANGE: `LdtkAsset` and `LdtkProject` are now exported in new `assets` module instead of `lib.rs`
BREAKING-CHANGE: asset `Loader` types are now private


Trouv:
Actually, decided to split the world_height stuff into its own PR here: https://github.com/Trouv/bevy_ecs_ldtk/pull/207

Issue #119, opened 2022-10-07
Use the default instance of LdtkEntity, not of its fields


HugoPeters1024:
I've run into the issue where it is not clear to me at which point I should define with what values I want to spawn my entities. For example,
I have a block blundle with a `SpriteSheetBundle`, a rapier 2d `Collider` and `RigidBody`, as well as a general marker struct:

```rust
#[derive(Bundle, LdtkEntity)]
struct BlockBundle {
    block: Block,

    body: RigidBody,

    collider: Collider,

    #[bundle]
    #[sprite_sheet_bundle]
    sprite: SpriteSheetBundle
}
```

I can very conveniently derive `LdtkEntity` because each individual field has the `Default` trait. However, should I want to configure the body to generally be `RigidBody::Fixed` it is not enough to simply implement a `Default` trait for the `BlockBundle` itself, since this is not considered by the deriving machinery of `LdtkEntity`. 

Therefore I was wondering if there is already an anticipated path that accomplishes what I want (other than creating an on create system that corrects these values, that feels fragile and wasteful to me) and if not, to suggest requiring a `Default` trait on the bundle itself and using that to derive `LdtkEntity`. If users do not which to alter the beheavior they can simple derive `Default` along with `LdtkEntity` and continue enjoying the semantics of the current situation.

Happy to hear anyone's thoughts and my compliments to all who contributed, it is a great project :)

(P.S. should there be consensus about implementing my suggestion, I'd be happy to give it a shot and gain some experience)

Trouv:
This is a good suggestion. Using the bundle's `Default` vs the `Default` of the individual components is not a decision that we should make lightly.

As for what you can do now - you can always manually implement `LdtkEntity` rather than using the derive impl. It's a bit verbose though due to the method signature, which is annoying.

For the sake of modularity, I find myself implementing `From<EntityInstance>` or `LdtkEntity` for individual components, and then giving them the `#[from_entity_instance]` or `#[ldtk_entity]` attributes respectively. However, I've never found this solution to be very ergonomic or enjoyable, as you technically should be checking the value of the entity's identifier in these manual implementations, especially if you intend to use that component in multiple bundles.

I don't really remember why I decided to use the `Default` impl of the individual fields rather than the `Default` impl of the bundle. I do remember explicitly making that decision though. There must have been some reason I went the way I did, but I probably didn't consider the value potential you've brought up here. I'll give this some more thought.

elenakrittik:
[...]

PR #222, opened 2023-08-19
feat!: More intuitive default instantiation behavior.

elenakrittik:
Changes the default behavior to use bundle's struct's `Default` `impl` instead of each individual's field.

Solves #119.

- [x] I have properly formatted code using `cargo fmt`.
- [x] I have tested the code and documentation i changed by both crafting and running an example and by running `cargo test`.
- [x] I have ran `cargo clippy` and fixed any issues arising from my changes.


Trouv (commented):
It may be better to just make this the normal behavior, not opt-in or opt-out. It's certainly more flexible to make it opt-in like this, but I think the current behavior is a little confusing and it would be fine to just get rid of it. Making this the only behavior would also be a bit easier to maintain/test in the future, rather than having one more toggle-like feature to worry about. Also, I've been considering actually scaling back the macro functionality in future releases (not soon, like 0.11 or something), and if that comes to pass the opt-inness of this would probably be on the chopping block anyway.

Down side would be that it's a breaking change, but I'm not too worried about that. This library is far from stable, and I'm planning to release 0.9 with many breaking changes with plenty of time before bevy 0.12 comes out so people can upgrade at their own pace. Thoughts?

elenakrittik:
I personally totally share your vision and if breaking changes are for the better, then i'll make them. But at the same time i recognize that there's a plenty of developers with busy schedules and that they would rather prefer to receive bug fixes without the accompanying breaking changes. Though, in case of Bevy's ecosystem, i'm imagine most people will be fine with breaking changes, or pin bevy-related dependencies to specific versions and upgrade only when they're ready. I'll change this PR a bit later.

elenakrittik:
Done. Due to removal of backwards compatibility had to add lots of `#[derive(Default)]`s.

macros/src/ldtk_entity.rs
```diff
@@ -96,10 +96,6 @@ pub fn expand_ldtk_entity_derive(ast: &syn::DeriveInput) -> proc_macro::TokenStr
             field_constructions.push(expand_with_attribute(attribute, field_name, field_type));
             continue;
         }
-
-        field_constructions.push(quote! {
-            #field_name: <#field_type as std::default::Default>::default(),
-        });
```
Trouv, line 102: Do you think it would be a good idea to make it so the `..Default::default()` is only added if there are any remaining fields? That would make so technically `LdtkEntity` and `LdtkIntCell` bundles don't *have* to implement default, so long as all their fields are constructed by some other method, so a little more flexible. However, I could also see that just being a little bit confusing.

examples/basic.rs
```diff
@@ -28,7 +28,7 @@ struct ComponentA;
 #[derive(Default, Component)]
 struct ComponentB;
 
-#[derive(Bundle, LdtkEntity)]
+#[derive(Default, Bundle, LdtkEntity)]
```
Trouv, line 31: Do you think you could make this change in the `README.md` as well?

src/app/ldtk_entity.rs
```diff
@@ -50,7 +50,7 @@ use std::{collections::HashMap, marker::PhantomData};
 /// # struct ComponentB;
 /// # #[derive(Component, Default)]
 /// # struct ComponentC;
-/// #[derive(Bundle, LdtkEntity)]
+/// #[derive(Bundle, LdtkEntity, Default)]
 /// pub struct MyBundle {
 ///     a: ComponentA,
 ///     b: ComponentB,
```
Trouv, line 56: On line 63, the docs say 
```rust
/// By default, each component or nested bundle in the bundle will be created using their [Default]
/// implementations.
```
It's a bit ambiguous but I think it's trying to describe the old behavior. I think we could make this a little more clear now by saying something like `By default, the fields of the bundle will be created using its [Default] implementation.`

src/app/ldtk_int_cell.rs
```diff
@@ -46,7 +46,7 @@ use std::{collections::HashMap, marker::PhantomData};
 /// # struct ComponentB;
 /// # #[derive(Component, Default)]
 /// # struct ComponentC;
-/// #[derive(Bundle, LdtkIntCell)]
+/// #[derive(Bundle, LdtkIntCell, Default)]
 /// pub struct MyBundle {
 ///     a: ComponentA,
 ///     b: ComponentB,
```
Trouv, line 52: Same comment about the docs on line 59

Trouv (changes requested):
Thanks for making the breaking change, I think it's a bit better this way

macros/src/ldtk_entity.rs
elenakrittik: Not worth it IMHO. Hopefully made the docs clear enough.

src/app/ldtk_int_cell.rs
elenakrittik: 8527b03

src/app/ldtk_entity.rs
elenakrittik: 8527b03

examples/basic.rs
elenakrittik: 8527b03

macros/src/ldtk_entity.rs
elenakrittik: Huh, seems latest clippy will force us to do this. Although i can't reproduce it locally.

Trouv: That or you can just allow the clippy lint in the macro. Either way is fine with me

elenakrittik: 2cbcbbe

elenakrittik:
Done, sorry for this simple change taking so long.

Trouv (approved):
> Done, sorry for this simple change taking so long.

No worries. We're all volunteers here. Thanks again for the contribution!

PR #223, opened 2023-08-19
fix: Fix clippy warnings introduced in rustc 1.72

elenakrittik:


Trouv (approved):
LGTM, thank you for making this a separate PR. I'm going to change the commit message to "chore", but that means it won't get credited in the changelong/release notes, sorry about that.

Issue #227, opened 2023-09-12
"overflow" layer_entities too large


MScottMcBee:
# Expected Result

Given a 10x10 int_grid named "map" with 10 tiles in one space and one tile in the remaining spaces, `spawn_level` should spawn 1 entity named "map" with 100 tiles, and 9 additional "map" entities each with 1 tile each to cover the "overflow"

# Observed Result

Given the same map as above, `spawn_level` will spawn 1 "map" entity with the correct 100 tiles, and then the 9 overflow map entities will also have 100 tiles each, 99 of which are invisible tiles. All the invisible tiles can take a big toll on performance and memory in larger levels.

# Suspected cause

`layer_grid_tiles` splits the tiles into correct `grid_tiles` Vecs,  `tile_pos_to_tile_if_int_grid_nonzero_maker` checks against the whole `int_grid_csv` array on where to create a tile, and not `grid_tiles`, so an invisible tile is created where ever a tile exists on any `layer_entity`

I've made a fix for this issue in my fork, and I've included a PR for it here: https://github.com/Trouv/bevy_ecs_ldtk/pull/228

MScottMcBee:
I spent some more time on this issue. I made a small level in LDTK to reproduce it, and it got messy quick. Default options in LDTK prevent the issue from happening in a lot of cases and #228 is broken in a few ways, mostly in how I set coordinates. In working through the issues, though, I've come to believe that creating invisible tiles for Auto-layer layers doesn't serve a purpose in the most common use of LDTK. 

I realize this opinion does have a decent amount of ignorance behind it (I'm not the one who's been working on this library for years), but I've submitted[ a PR for consideration anyway](https://github.com/Trouv/bevy_ecs_ldtk/pull/230). Feel free to disregard.

Trouv:
To elaborate on the purpose of invisible tiles in general - it's pretty much exclusively to ensure that intgrid cells will spawn appropriately for intgrid+autotile layers. Intgrid values tend to be pretty important to game logic and the APIs provided by this plugin. So, on intgrid+autotile layers, tiles with a nonzero intgrid value need to exist whether or not the autotile rules of that layer determine that they have a tileset value. Conversely, the tile needs to exist (and be visible) on intgrid+autotile layers if it has a tileset value but happens to have a 0 intgrid value. 

[...]

PR #228, opened 2023-09-12
tile_pos_to_int_grid_with_grid_tiles_tile_maker checks grid_tiles

MScottMcBee:
A possible solution to https://github.com/Trouv/bevy_ecs_ldtk/issues/227, this PR changes `tile_pos_to_tile_if_int_grid_nonzero_maker` to take in an `grid_tiles` array as an argument and checks against it to see if an invisible tile should be made or not.

### DIFF
--- a/src/level.rs
+++ b/src/level.rs
@@ -524,9 +524,9 @@ pub fn spawn_level(
                                                         tile_pos_to_tile_if_int_grid_nonzero_maker(
                                                             tile_pos_to_invisible_tile,
                                                             &layer_instance.int_grid_csv,
+                                                            &grid_tiles,
                                                             layer_instance.c_wid,
                                                             layer_instance.c_hei,
-                                                            None
                                                         ),
                                                         layer_instance.opacity,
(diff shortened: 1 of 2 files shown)


src/level.rs
```diff
@@ -524,6 +524,7 @@ pub fn spawn_level(
                                                         tile_pos_to_tile_if_int_grid_nonzero_maker(
                                                             tile_pos_to_invisible_tile,
                                                             &layer_instance.int_grid_csv,
+                                                            &grid_tiles,
```
Trouv, line 527: So I'm a little confused by this actually - the issue is for grid_tiles overflow, right? I thought that could only happen on AutoTile layers (and intgrid+autotile layers). However, this change only affects non-autotile intgrid layers. The `match tileset_definition { ... }` this is inside of is `Some` on autotile intgrids, and `None` on non-autotile intgrids.

Trouv:
Thanks for the detailed issue, and the contribution!

PR #231, opened 2023-09-19
Don't spawn invisible tiles on sublayers other than the first

MScottMcBee:
Another possible solution to #227. Based off of the previous settings PR, this changes it so that instead of passing a setting to spawn invisible tiles or not, it passes the sublayer index.


Trouv (commented):
Perfect. Thanks for taking care of this and doing some rewrites and updating internal docs/tests before I even had to ask

PR #234, opened 2023-10-05
Fix incorrect use of LevelSelection in example.

jaapieaapie1:
A very small change but I noticed that the example had a typo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ fn main() {
         .add_plugins(DefaultPlugins)
         .add_plugins(LdtkPlugin)
         .add_system(Startup, setup)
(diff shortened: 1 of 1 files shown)


Trouv:
Sorry, this is actually intended. There have been many breaking changes merged to main since `bevy_ecs_ldtk` 0.8 and the examples have been updated accordingly. One such change is that `LevelSelection::Index(usize)` has been replaced with `LevelSelection::Indices(LevelIndices)` to better support multiple-world projects. And, to help maintain the old functionality somewhat, a method has been added called `LevelSelection::index` that constructs this variant with a `None` world index.

PR #244, opened 2023-10-16
feat!: redesign LdtkProject with better level data accessors and correct modeling of internal/external levels

### LINKED ISSUE #205: Provide API abstracting over external/internal levels better
Currently, `LdtkLevel` is the de-facto abstraction for retrieving level data in both external and internal cases. A `Handle<LdtkLevel>` is loaded as a labeled asset in the internal case, and it is loaded as a dependent asset in the external case. However, Since `LdtkAsset` already has this level data in the internal case, it ends up being cloned in order to create an `LdtkLevel`. Surely there's a way to leverage a smart pointer and/or enums to avoid this expensive clone.

Something to the effect of...
```rust
enum LevelPointer {
    Internal {
        data: Arc<Level>,
        bg_image: Handle<Image>,
    },
    External {
        handle: Handle<LdtkExternalLevel>
    },
}
```

Trouv:
Closes #205

This is the final PR for redesigning the asset types. The main improvements of this redesign are that the type provides better APIs for accessing raw or loaded level data, and internal/external levels are modeled more correctly.

The `LdtkProject` type is still the asset type and stores most metadata applicable to either level locale, along with `LdtkProjectData`.

`LdtkProjectData` is a enum whose variants store the actual ldtk json data, with a variant for each level locality. The internal types of the variants are `LdtkJsonWithMetadata<L: LevelLocale>`, with either `InternalLevels` or `ExternalLevels` as `L`.

`LdtkJsonWithMetadata<L>` is a generic type storing the actual ldtk json data + level metadata, with either `InternalLevels` or `ExternalLevels` as `L`.

Splitting these up like this allows us to define some methods that are exclusive to each level locality. This is important because accessing loaded level data is a very different operation memory-wise for each case (indexing the `LdtkJson` for internal-levels, or accessing the `Assets<LdtkLevel>` asset store for external levels). But other methods can be provided for either case, either with a generic implementation at the lowest level, or exposing transparent implementations in the higher-level types.

An important point about this new design is that `LdtkLevel` assets are no longer used in the internal-levels case. Level entities will only ever have a `LevelIid` component, no longer a `Handle<LdtkLevel>`. The handle for the asset in the external-levels case is only stored inside the `LdtkProject`. To help make this change clear, `LdtkLevel` has been renamed `LdtkExternalLevel`. Doing things this way actually fixes an egregious clone of all level data that is in all previous versions of this plugin.

See the documentation of `LdtkProject` to learn about the best ways to access loaded level data now.

feat!: LdtkLevel renamed to LdtkExternalLevel and is no longer used as a component (#244)

### DIFF
--- a/src/assets/ldtk_external_level.rs
+++ b/src/assets/ldtk_external_level.rs
@@ -22,6 +22,11 @@ pub struct LdtkExternalLevel {
 }
 
 impl LdtkExternalLevel {
+    #[cfg(test)]
+    pub fn new(data: Level) -> LdtkExternalLevel {
+        LdtkExternalLevel { data }
+    }
+
     pub fn data(&self) -> LoadedLevel {
         LoadedLevel::try_from(&self.data)
             .expect("construction of LdtkExternalLevel should guarantee that the level is loaded.")
--- a/src/assets/ldtk_external_level.rs
+++ b/src/assets/ldtk_external_level.rs
@@ -17,28 +17,36 @@ use thiserror::Error;
 #[derive(Clone, Debug, PartialEq, TypeUuid, Reflect)]
 #[uuid = "5448469b-2134-44f5-a86c-a7b829f70a0c"]
 pub struct LdtkExternalLevel {
-    /// Raw ldtk level data.
+    /// Raw LDtk level data.
     data: Level,
 }
 
 impl LdtkExternalLevel {
+    /// Construct a new [`LdtkExternalLevel`].
+    ///
+    /// Only available for testing.
+    /// This type should only be constructed via the bevy asset system under normal use.
     #[cfg(test)]
     pub fn new(data: Level) -> LdtkExternalLevel {
         LdtkExternalLevel { data }
     }
 
+    /// Internal LDtk level data as a [`LoadedLevel`].
     pub fn data(&self) -> LoadedLevel {
         LoadedLevel::try_from(&self.data)
             .expect("construction of LdtkExternalLevel should guarantee that the level is loaded.")
     }
 }
 
+/// Errors that can occur when loading an [`LdtkExternalLevel`] asset.
 #[derive(Debug, Error)]
 pub enum LdtkExternalLevelLoaderError {
-    #[error("external LDtk level should contain all level data, but the level's layers is null")]
+    /// External LDtk level should contain all level data, but some level has null layers.
+    #[error("external LDtk level should contain all level data, but some level has null layers")]
     NullLayers,
 }
 
+/// AssetLoader for [`LdtkExternalLevel`]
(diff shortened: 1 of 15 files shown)


src/assets/ldtk_project.rs
```diff
@@ -41,22 +139,98 @@ impl RawLevelAccessor for LdtkProject {
...
+    /// LDtk project uses external levels, but some level's `external_rel_path` is null.
+    #[error("LDtk project uses external levels, but some level's external_rel_path is null")]
+    ExternalLevelWithNullPath,
+}
+
+/// AssetLoader for [`LdtkProject`].
 #[derive(Default)]
 pub struct LdtkProjectLoader;
 
+struct LoadLevelMetadataResult<'a, L> {
```
Trouv, line 170: NIT: could use LevelLocale as a trait bound here

src/systems.rs
```diff
@@ -4,55 +4,58 @@
...
                 debug!("LDtk asset creation detected.");
-                ldtk_handles_for_clear_color.insert(handle);
+                ldtk_handles_for_clear_color.insert(handle.clone());
             }
             AssetEvent::Modified { handle } => {
                 info!("LDtk asset modification detected.");
-                ldtk_handles_to_respawn.insert(handle);
-                ldtk_handles_for_clear_color.insert(handle);
+                ldtk_handles_to_respawn.insert(handle.clone());
+                ldtk_handles_for_clear_color.insert(handle.clone());
```
Trouv, line 43: Are these clones necessary?

Trouv:
FYI @Hellzbellz123 this may affect you. `Reflect` is still only implemented for the level asset type, but now those assets are only created if the project uses external levels

src/assets/ldtk_project.rs
Trouv: not really worth it as it requires removing `#[cfg(feature = "internal_levels")]` from `InternalLevels`

src/assets/ldtk_external_level.rs
```diff
@@ -0,0 +1,74 @@
...
+
+impl LdtkExternalLevel {
+    #[cfg(test)]
+    pub fn new(data: Level) -> LdtkExternalLevel {
+        LdtkExternalLevel { data }
+    }
+
+    pub fn data(&self) -> LoadedLevel {
+        LoadedLevel::try_from(&self.data)
+            .expect("construction of LdtkExternalLevel should guarantee that the level is loaded.")
```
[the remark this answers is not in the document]
Trouv: Getters doesn't make sense here since `.data()` coerces to a `LoadedLevel` rather than just handing back a plain ref. However, everything else is done.

PR #255, opened 2023-10-21
Normalize resolved asset paths using `path_clean`.

### LINKED ISSUE #240: [Feat] Normalize paths or allow asset path customization
I'm using [bevy_asset_loader](https://github.com/NiklasEi/bevy_asset_loader) to preload all assets, and currently keeping LDtk files in `assets/levels` adjacent to other paths like `assets/spritesheets`. The reasons for the latter aren't absolutely critical, but (a) multi-worlds feature is still considered experimental, (b) I'd like to support "external" mod-worlds, and (c) it helps to keep things organized in general.

Anyway, while working on a feature today that depends on shared assets being... well, _shared_, I discovered that `bevy_ecs_ldtk` was creating a second copy of all the assets referenced in the LDtk project. This is because the LDtk paths are relative to the project file, and when passed to the asset loader, are in the form of `levels\\../spritesheets/foo.png`. These paths do load, of course, but Bevy's asset loader assigns them new `AssetPathId`s despite pointing to the same file as `spritesheets/foo.png`, so they get loaded and added as new assets.

I realize that _in theory_ there's a lot of madness around symlinks, and other very good reasons why all the pieces involved behave as they do. Nevertheless, in practice, and certainly in my project, those edge cases are never going to happen, and what I do care about is making sure that the LDtk systems and non-LDtk systems get the same asset when they request the same asset.

Slapping a [path-clean](https://github.com/danreeves/path-clean) on the `ldtk_path_to_asset_path` method solves the problem beautifully for me, but requires a change directly to the source code; I can't intercept this in the app code. Perhaps it is too blunt an instrument to apply to every user - maybe some people really do care about nuances like symlinks - so if that is the case, then I'd like to request canonicalization as an opt-in feature, whether it's something simple like a flag in `LdtkSettings` or requires us to do more of the legwork e.g. by implementing some trait with the actual conversion function.

It's probably a weird edge case and I don't expect it to get high priority, given that the workaround is technically trivial (just make sure all LDtk projects are in the asset root), but wanted to bring it up particularly because there's no warning that it's happening. My project is still small, but in a project with hundreds of MBs of assets, it could end up doubling the loading time and wasting a lot of memory. Also, `bevy_ecs_ldtk` having dedicated code to resolve LDtk asset paths suggests that it's designed to accommodate these scenarios, and does handle them, but won't play nice with other systems unless the paths are normalized.

focustense:
Makes all paths relative to the `assets` root. Ensures that LDtk-linked assets are shared with assets loaded by `AssetServer`, [bevy_asset_loader](https://github.com/NiklasEi/bevy_asset_loader) and so on, as long as normalized paths are used for those (as they typically are), especially when the LDtk project is in a subdirectory.

Fixes #240

### DIFF
--- a/src/assets/ldtk_project.rs
+++ b/src/assets/ldtk_project.rs
@@ -393,6 +393,7 @@ mod tests {
     }
 
     #[cfg(target_os = "windows")]
+    #[test]
     fn normalizes_windows_asset_paths() {
         let resolve_path = |project_path, rel_path| {
             let asset_path = ldtk_path_to_asset_path(Path::new(project_path), rel_path);
--- a/src/assets/ldtk_project.rs
+++ b/src/assets/ldtk_project.rs
@@ -14,8 +14,8 @@ use bevy::{
 };
 use derive_getters::Getters;
 use derive_more::From;
-use std::collections::HashMap;
 use path_clean::PathClean;
+use std::collections::HashMap;
 use thiserror::Error;
 
 #[cfg(feature = "internal_levels")]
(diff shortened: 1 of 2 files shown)


focustense:
The failing test looks like a platform-specificity issue; the native `Path` must treat the backslash as just a regular character, so `path_clean` ignores it too. I'll look into fix tomorrow - probably just make that test case Windows-conditional.

src/assets/ldtk_project.rs
```diff
@@ -365,6 +371,48 @@ mod tests {
...
+            Path::new("projects/images/tiles.png")
+        );
+        assert_eq!(
+            resolve_path("projects/sub/project.ldtk", "../../tiles.png"),
+            Path::new("tiles.png")
+        );
+    }
+
+    #[cfg(target_os = "windows")]
+    fn normalizes_windows_asset_paths() {
```
Trouv, line 396: This isn't decorated w/ `#[test]` so I don't think it would run on windows either.

I'm also okay leaving this test out entirely, the CI does not build on windows and neither do I so it would be completely invisible in future reviews.

Trouv (changes requested):

src/assets/ldtk_project.rs
focustense: Derp, rookie mistake there. Added the attribute, it still runs and passes fine - on my machine, anyway.

Maybe there's a chance it'll get stale under current maintenance conditions, but I think it's better to keep the test anyway. This code rarely seems to change, and _someone_ will eventually notice if it breaks on Windows and if so, the test will tell them why.

If it's important to you that the test be removed then let me know and I can do that. Otherwise, should be resolved.

Trouv (approved):
LGTM. Thanks for the suggestion and the tests as well

PR #275, opened 2023-12-03
Add exclusions to `LdtkSettings` and filters to level spawner.

### LINKED ISSUE #272: [feat] Add an option to disable spawning of some/all IntGrid entities
I've added a few different IntGrids to track different types of tile metadata and noticed that the result is essentially a new entity hierarchy, tilemap, etc. for each grid under the level entity.

I appreciate that we can make the IntGrid tiles invisible, but in my particular case, I really have no use for these entities at all, and I'm a little concerned about their effect on performance. I'm not using them for AutoLayers, I don't need to see them in the game world (because I can see them very easily in LDtk itself), and there is a _lot_ of redundant data being spawned - hundreds of entities each with dozens of components, many of which are going to trigger some tilemap systems and other systems to run despite being invisible to the eye. It's compounded when I'm loading up to 4 level neighbors.

Would it be possible to add something to the `LdtkSettings`, or wherever makes the most sense, that allows either ignoring _all_ IntGrid layers (except AutoLayers, I guess) or selectively whitelisting/blacklisting the layers we're interested/not interested in?

I'd like to also point out that when using IntGrid as metadata - which is kind of the point, AutoLayers weren't implemented until much later - it is usually _not_ faster to query the ECS than it is to just query the level data in the LDtk project, because the ECS is not indexed by any useful key. Whereas, in the LDtk project itself, if we know the grid coordinates of the tile we want to query, and also know the grid size, then it is simple and cheap math to translate that into an index and no problem at all to do this a few hundred times per frame.

(P.S. I totally get that it will be useful for _some_ users to get a whole entity tree for those layers, maybe they want to run additional systems or spawn new entities based on it, etc., and those _are_ better optimized in ECS style. That's why I'm asking for it to be a choice.)

(P.P.S. If you think it's a useful request, but don't have time/are busy with other priorities, I'm willing to take a crack at a PR if you could give me a few pointers re: where the magic happens.)

focustense:
I made this a separate `LdtkExclusions` struct because, even though there's only one use case for this now, it would be a logical extension to add other exclusions in the future (like entities, both IID and UID based), and if so, it's better to group them together instead of cluttering up the main `LdtkSettings`.

Caveat: Adding a `Vec` (or any other collection) to the `LdtkSettings` or any sub-structure means we can no longer derive `Copy`. It doesn't appear to be required by any part of `bevy_ecs_ldtk` itself, and there's no compelling reason why users should absolutely need it (copying a `Resource` is pretty rare?), but this is potentially a breaking change for some users. The only way around this would be to avoid putting this in `LdtkSettings` altogether and define the `LdtkExclusions` as its own resource (and modify the various systems accordingly). I feel that's a YAGNI problem so I went the easy route and just removed `Copy`.

Unit testing seems infeasible at the moment due to the complexity of the `spawn_level` method and absence of existing test infra for it, but I manually tested the exact same change on the 0.8 branch in an actual game and it worked perfectly.

Note also that the use of `Vec` over `HashSet` is intentional. I'd expect there to be a very small number of items in that list (say 2 or 3). In lists with fewer than 10-15 items, the cost of the hash tends to exceed the cost of the iteration ([example](https://gist.github.com/daboross/976978d8200caf86e02acb6805961195)).

Fixes #272

### DIFF
--- a/src/resources/mod.rs
+++ b/src/resources/mod.rs
@@ -2,6 +2,8 @@
 use bevy::prelude::*;
 
 #[allow(unused_imports)]
+use crate::assets::LdtkProject;
+#[allow(unused_imports)]
 use crate::components::LdtkWorldBundle;
 
 mod level_selection;
@@ -67,12 +69,21 @@ pub enum LevelBackground {
     Nonexistent,
 }
 
+/// Specifies data that should be ignored completely when spawning levels. Excluded items will still
+/// be present in the [`LdtkProject`] but will not cause any entities to be spawned in the world.
+#[derive(Clone, Eq, PartialEq, Debug, Default)]
+pub struct LdtkExclusions {
+    /// List of layer `Identifier` names (not UIDs) to skip spawning as tilemaps.
+    pub layer_identifiers: Vec<String>,
+}
+
 /// Settings resource for the plugin.
 /// Check out the documentation for each field type to learn more.
-#[derive(Copy, Clone, Eq, PartialEq, Debug, Default, Resource)]
+#[derive(Clone, Eq, PartialEq, Debug, Default, Resource)]
 pub struct LdtkSettings {
     pub level_spawn_behavior: LevelSpawnBehavior,
     pub set_clear_color: SetClearColor,
     pub int_grid_rendering: IntGridRendering,
     pub level_background: LevelBackground,
+    pub exclusions: LdtkExclusions,
 }
--- a/src/resources/mod.rs
+++ b/src/resources/mod.rs
@@ -72,7 +72,7 @@ pub enum LevelBackground {
 /// Specifies data that should be ignored completely when spawning levels. Excluded items will still
 /// be present in the [`LdtkProject`] but will not cause any entities to be spawned in the world.
 #[derive(Clone, Eq, PartialEq, Debug, Default)]
-pub struct LdtkExclusions {
+pub struct SpawnExclusions {
     /// List of layer `Identifier` names (not UIDs) to skip spawning as tilemaps.
     pub layer_identifiers: Vec<String>,
 }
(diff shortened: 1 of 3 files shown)


Trouv:
Thanks for the contribution!

> I made this a separate LdtkExclusions struct because, even though there's only one use case for this now, it would be a logical extension to add other exclusions in the future

Nice, this is definitely an improvement over my original suggestion

> I feel that's a YAGNI problem so I went the easy route and just removed Copy.

agreed

> Unit testing seems infeasible at the moment due to the complexity of the spawn_level method and absence of existing test infra for it

That's fine. I plan to rewrite this plugin's systems to be smaller/more modular eventually, and when that time comes I will have tests for them and begin asking contributors for tests as well. For now they're definitely unreasonably difficult to test.

src/resources/mod.rs
```diff
@@ -67,12 +69,21 @@ pub enum LevelBackground {
     Nonexistent,
 }
 
+/// Specifies data that should be ignored completely when spawning levels. Excluded items will still
+/// be present in the [`LdtkProject`] but will not cause any entities to be spawned in the world.
+#[derive(Clone, Eq, PartialEq, Debug, Default)]
+pub struct LdtkExclusions {
```
Trouv, line 75: I think maybe something like `SpawnExclusions` might be slightly more descriptive. Also `Ldtk` part of the name isn't super helpful at this point in the library because it's already in the `LdtkSettings` resource

Trouv, line 75: Do you think you could re-export this in the prelude?

Trouv (changes requested):
A couple minor notes

src/resources/mod.rs
focustense: Done.

focustense: Done.

Trouv (approved):

PR #280, opened 2023-12-13
ci: publish book to versioned subdirectories

### LINKED ISSUE #274: Book versioning
Currently, the book is being published to github pages on any new version tags pushed to the repo. This is a good starting point for automatic releases, but ideally there should be some versioning mechanism in the resulting website.
Two big use cases for this
- users of older versions of the plugin should be able to read the corresponding version of the book
- it's handy to be able to preview newer versions of the book in the actual environment it will be published in as the plugin/book develops

Since github pages uses static files for rendering, I don't think this is too unreasonable. In the pages branch, each version-tag-publish of the book could produce a new directory named after the version. Then, each push to main could publish the book to a `main` directory in the pages branch. A `latest` directory could redirect to the most-recently-published version-tag of the book. Then, the index.html of the entire pages branch could either have a list of all of these, or just redirect to latest.

This seems to have been accomplished before, and we may be able to use existing technologies like https://github.com/koumoul-dev/gh-pages-multi

Ideally, the other versions of the book would somehow be discoverable from each version of the book.

Trouv:
Closes #274 

This changes the publish book CI to subdirectories of the github pages branch rather than the root. The directory name is either the tag name that triggered the publishing, or the branch name (which can only be `main` due to the workflow triggers). Furthermore, if it is a tag that triggered the publishing, the `latest` directory is *also* published to. So the result will be multiple versions of the book published to github pages, one for each version of the plugin named after that version, plus two special versions: `latest` and `main`. This will help with development of the book and allow users of older versions of the plugin to reference older documentation.

### DIFF
--- a/.github/workflows/publish-book.yml
+++ b/.github/workflows/publish-book.yml
@@ -2,6 +2,8 @@ name: publish book
 
 on:
   push:
+    branches:
+      - main
     tags:
       - 'v*.*.*'
 
--- a/.github/workflows/publish-book.yml
+++ b/.github/workflows/publish-book.yml
@@ -38,3 +38,4 @@ jobs:
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: ./book/book
+          destination_dir: ./${{ github.ref_name }}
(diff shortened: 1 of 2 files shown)


Trouv:
https://trouv.github.io/bevy_ecs_ldtk now redirects to https://trouv.github.io/bevy_ecs_ldtk/latest

PR #281, opened 2023-12-14
Don't set level selection if the level isn't loaded yet

### LINKED ISSUE #232: Race condition in external level projects
I have an LDTK project using external levels that has 34 levels. When I load a project file and select a level, `process_ldtk_levels` will sometimes fail to load the level. Inspecting the size of `Assets<LdtkLevel>` will show between 22 and 31.

What seems to be happening is that when the `LdtkProject` asset gets loaded it creates all the handles for the external levels, but [loading asset dependencies isn't guaranteed to happen all at once](https://discord.com/channels/691052431525675048/1124009016062918666/1152742978918547616). Since `process_ldtk_levels` is triggered by the creation of the handle, it's possible that your level just isn't there.

I was able to make a rough fix on my fork [here](https://github.com/MScottMcBee/bevy_ecs_ldtk/pull/2/files) by letting `process_ldtk_levels` run until the number of assets loaded matches the number of external files listed in a project.

Seeing as how Assets v2 is already in bevy's main branch, this might not be worth fixing at this exact moment.

MScottMcBee:
Fixes #232 

Added a check so that when applying `LevelSelection`, there's a check to make sure the external level is loaded.

Works for my game and all tests pass, not sure if this has any weird knock-on effects.


src/systems.rs
```diff
@@ -74,12 +74,27 @@ pub fn apply_level_selection(
...
+                            let loaded_level = project_data.get_external_level_at_indices(
+                                &level_assets,
+                                level_metadata.indices(),
+                            );
+
+                            if loaded_level.is_none(){
+                                return;
+                            }
+                        }
+                    }
```
Trouv, line 97: So I believe the way this currently works is that `apply_level_set` wraps everything in a `if let Some(project)` for the `LdtkProject`, which acts as a check that the project is loaded. We don't actually listen for `AssetEvent::Created` or `AssetEvent::LoadedWithDependencies` for this. This of course does *not* act as a check that levels are loaded, just that the project is loaded, which is the bug. I think you should be able to use the new [`AssetServer::get_recursive_dependency_load_state`](https://docs.rs/bevy/latest/bevy/asset/struct.AssetServer.html#method.get_recursive_dependency_load_state) method for a more correct check that the project (and its dependencies) are loaded.

Trouv (changes requested):
Thanks for the contribution and for following up on this issue. I see why the change in the 0.12 branch doesn't fix this now.

MScottMcBee:
Thanks for the feedback and sorry for the long response time. I switched over to use `get_recursive_dependency_load_state` and wow is that a lot cleaner.

Trouv (approved):
Seems good. Will merge this *after* `feat/bevy-0.12` so that the squash commit credits you properly.

Issue #177, opened 2023-04-07
Proposal for "bundler systems" and macros v2


Trouv:
# Motivation
I've been dissatisfied with the `LdtkEntity` and `LdtkIntCell` traits/derive macros for a while now. In this issue, I will mostly be focusing on `LdtkEntity` since it is the more complex of the two, and pretty much all of these changes should be applied to each. I go into detail about some of my complaints in #47. Basically, my complaints are..
- the available macros/attributes are too opinionated about how the users will want to construct their bundles
- the manual implementation of these traits is also very opinionated about what information the user may need access to
  - for simple manual implementations of `LdtkEntity`, most method parameters go unused and add a lot of boilerplate
  - for more complex manual implementations of these traits, the method parameters may not be enough
- If users want to construct the same type in different ways for multiple registrations, the plugin encourages some bad designs such as..
  - For each registration, a different singleton bundle with a different manual `LdtkEntity` implementation
  - All-encompassing `LdtkEntity` implementations that manually check the entity's `identifier` internally to differentiate
  - Similar designs for the components/nested bundles, but using `From<&EntityInstance>` instead

I think the final point is partially solved by the relatively-new `#[with]` attribute macro. This allows you to provide a custom constructor to a field in the bundle, without being bound to a trait. This is perfect for situations where you want to construct a component in two different ways for two different registrations. However, it is limited because your constructor can only accept an `&EntityInstance`. Even if we gave it more arguments, it would still have a similar issue to the `LdtkEntity` trait itself - we're still being opinionated about what information the user may need for construction.

The solution to the issue of "opinionated dependencies" is dependency injection. This is what I was trying to express in #47. If we could somehow allow users to provide constructors that are more like bevy systems, where they define in the function arguments what information they need from the world, many of these issues go away. I liked the idea of re-using actual bevy systems, but there were many design questions around this, and I wasn't totally sure that it was possible. For example, I wasn't sure how to best provide metadata pertaining to the entity outside of system parameters (like `&EntityInstance` or `&LayerMetadata`).

# Proposal Summary
[...]

PR #294, opened 2024-01-16
Fix ignored entity pivot

### LINKED ISSUE #290: Loaded entity position ignores pivot
In my LDtk project, I have defined a number of entities with a bottom pivot point: 
image

Based on my work in other apps, I expect the pivot to change how entities are placed relative to the position shown--and indeed, in LDtk I see this exact thing happening.
image
_Left: origin on the bottom. Right: origin in center, moved to fill the same space. Note the changed Y coordinate._

But when those entities are loaded into Bevy:
image
What the pfargtl?! The `Transform` component stores the object's _center_ instead of its origin! 

I've compensated in my code. But it's a huge bother to have to do complicated math instead of just reading a component which, frankly, ought to match LDtk's behavior in the first place (apart from Y-down coordinates).

Koopa1018:
This PR modifies `calculate_transform_from_entity_instance` to use plain `ldtk_pixel_coords_to_translation` rather than the `pivoted` version. It also updates the function's test suite to reflect the different results it now gives.

Theoretically this PR will fix #290, but I've never worked with this codebase, so I have no idea if what I've done here is enough to actually fix the bug, nor do I know what else may need updating. (I know the enemies in the Platformer example show up displaced now.)


src/utils.rs
```diff
@@ -85,11 +85,9 @@ pub fn calculate_transform_from_entity_instance(
 
     let size = IVec2::new(entity_instance.width, entity_instance.height);
 
-    let translation = ldtk_pixel_coords_to_translation_pivoted(
+    let translation = ldtk_pixel_coords_to_translation(
```
Trouv, line 88: This is definitely a part of the change that needs to happen, to merge this though this PR should replace this with the anchoring functionality. I believe this will require updating the `TextureAtlasSprite` construction in `sprite_sheet_bundle_from_entity_info` to use the `pivot` value of `&EntityInstance`

Trouv (changes requested):
Thanks for the contribution!

> nor do I know what else may need updating

No worries. The examples definitely should be updated to match their previous behavior.

Also, since this is a breaking change, a new section in the migration guide is in order (at `book/src/how-to-guides/migration-guides/migrate-from-0.8-to-0.9.md`)

src/utils.rs
Koopa1018: Ahhh, in the sprite bundle...that makes sense. Glad you reminded me, that would've been a stupid thing to miss 🤦

Koopa1018:
Arright, I've done what I can with the examples. Most appear fine, and `traitless` was easy enough to fix, but I have no idea where to even look in the platformer example....
My only leads are that when the example begins running, the skulls slowly rise into the air as they move, until they reach this height: 
image
and that the boss (who does not move) has a lowered hitbox:
image

Migration guide is going smoothly. Not totally sure how to demonstrate the changed translation/pivot behavior in a code example, though....

Also: I notice the CI format check failed. I've never contributed to a project with CI, so I'm not sure what that means, even after checking the details. Are these formatting issues that I need to fix?

Trouv:
> Also: I notice the CI format check failed. I've never contributed to a project with CI, so I'm not sure what that means, even after checking the details. Are these formatting issues that I need to fix?

Yeah it'll need to be fixed before this is merged. For the format check in particular, it's just a matter of running `rustfmt`: https://github.com/rust-lang/rustfmt

Though it's something small enough that I'd do it myself if that was the only issue preventing merge.

BTW, some unsolicited advice: I recommend setting up your editor to automatically run rustfmt on save. It's so handy writing a ton of dirty code then hitting save and watching it prettify before your eyes

Koopa1018:
Can I get some help with the platformer example? It seems like to get the colliders to spawn right, it would require wider-reaching redesigns than I signed on for (that or I'm just totally missing a property that would fix everything...).

book/src/how-to-guides/migration-guides/migrate-from-0.8-to-0.9.md
```diff
@@ -429,3 +429,43 @@ let level_set = LevelSet::from_iids(
 );
 # }
 ```
+
+## Entity translations now respect pivot
+Entities now spawn with a translation matching their pivot, rather than their center point.
```
Trouv, line 434: I think implying that the new translation is their "actual position" whereas the old behavior is their "visual center" would be a little more clear
```suggestion
Entities now spawn with a translation matching their actual location, rather than their "visual center".
```

book/src/how-to-guides/migration-guides/migrate-from-0.8-to-0.9.md
```diff
@@ -429,3 +429,43 @@ let level_set = LevelSet::from_iids(
...
+```rust,ignore
+// How do I demonstrate that positions differ between old and new?
+// Must define pivot first.
+// Demonstrate loading an entity in 0.8, assert the position.
+// Then, do the same in 0.9, demonstrating that it's not the same.
+```
+The entities' sprites calculate their `Anchor` from the pivot, so they should appear the same,
```
Trouv, line 441: This is good but it should qualify this w/ the fact that this is only for sprite sheet bundles generated w/ the `#[sprite_sheet_bundle(...)]` attribute of `LdtkEntity`. Something like...
```suggestion
For `LdtkEntity` bundles with a `#[sprite_sheet_bundle(...)]`,  the macro calculates the sprite's `Anchor` from the pivot, so they should appear the same,
```

book/src/how-to-guides/migration-guides/migrate-from-0.8-to-0.9.md
```diff
@@ -429,3 +429,43 @@ let level_set = LevelSet::from_iids(
...
+		let object_center = crate::utils::entity_center(
+			transform.translation(),
+			IVec2::new(entity.width, entity.height),
+			entity.pivot
+		);
+
+		//do something with the center
+	}
+}
+```
```
Trouv, line 471: Yeah it's hard to say what the best transition strategy is. I think for things like rapier colliders. most users should probably make the collider a "child" of their entity rather than the main entity. Or, if they don't want to deal with that, they could make the entity's pivot center.

This goes for the platformer example as well.. but it's a big change so I completely understand if you're not interested in doing all that work. Don't worry, wherever you feel like leaving off on this PR I can pick it up (thanks for making it maintainer-editable)

examples/traitless.rs
```diff
@@ -51,6 +51,10 @@ fn process_my_entity(
 
                 let sprite = TextureAtlasSprite {
                     index: (tile.y / tile.h) as usize * 16 + (tile.x / tile.w) as usize,
+                    anchor: bevy::sprite::Anchor::Custom(Vec2::new(
+                        entity_instance.pivot.x - 0.5,
+                        0.5 - entity_instance.pivot.x,
```
Trouv, line 56: Ah this is interesting. Maybe having a utility like `fn ldtk_pivot_to_anchor(pivot: Vec2) -> Anchor` would be helpful.

src/utils.rs
```diff
@@ -227,13 +222,26 @@ pub fn ldtk_pixel_coords_to_translation_pivoted(
...
     let adjusted_pivot = Vec2::new(0.5 - pivot.x, pivot.y - 0.5);
 
+    // Find the spot on the entity which the pivot corresponds to.
     let offset = entity_size.as_vec2() * adjusted_pivot;
 
-    pivot_point + offset
+    translation + offset
+}
+
+pub fn entity_center(translation: Vec3, entity_size: IVec2, pivot: Vec2) -> Vec3 {
```
Trouv, line 237: I think `ldtk_entity_visual_center` could be a more consistent name for this.

Trouv, line 237: Could you leave a brief doc for this? Something like...
```suggestion
/// Calculate the "visual center" of an  LDtk object using its current bevy translation, size, and pivot value.
pub fn entity_center(translation: Vec3, entity_size: IVec2, pivot: Vec2) -> Vec3 {
```

src/utils.rs
```diff
@@ -227,13 +222,26 @@ pub fn ldtk_pixel_coords_to_translation_pivoted(
...
+    // Adjust entity pivot so that 0.5 is the center.
+    let adjusted_pivot = Vec2::new(0.5 - pivot.x, pivot.y - 0.5);
+
+    // Find the spot on the entity which the pivot corresponds to.
+    let offset = entity_size.as_vec2() * adjusted_pivot;
+
+    translation + offset.extend(0.0)
 }
```
Trouv, line 245: Also, some simple test cases for this would be amazing

Trouv (commented):
Thanks again for your work on this. The scope has definitely increased a lot, sorry about that.

One of the comments in this review does address a particular strategy for updating the platformer example. However, it's a lot of work, so I don't expect you to do it necessarily. I can happily pick up wherever you leave off.

examples/traitless.rs
```diff
@@ -51,6 +51,10 @@ fn process_my_entity(
 
                 let sprite = TextureAtlasSprite {
                     index: (tile.y / tile.h) as usize * 16 + (tile.x / tile.w) as usize,
+					anchor: bevy::sprite::Anchor::Custom(Vec2::new(
+						entity_instance.pivot.x - 0.5,
+						0.5 - entity_instance.pivot.x,
```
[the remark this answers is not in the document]
Koopa1018: Oh frick yes it absolutely does 🤦

examples/traitless.rs
Koopa1018: Ooh, good thought.

book/src/how-to-guides/migration-guides/migrate-from-0.8-to-0.9.md
```diff
@@ -429,3 +429,43 @@ let level_set = LevelSet::from_iids(
...
 ```
+
+## Entity translations now respect pivot
+Entities now spawn with a translation matching their pivot, rather than their center point.
+```rust,ignore
+// How do I demonstrate that positions differ between old and new?
+// Must define pivot first.
+// Demonstrate loading an entity in 0.8, assert the position.
+// Then, do the same in 0.9, demonstrating that it's not the same.
+```
```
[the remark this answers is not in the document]
Koopa1018: Screenshots! That's perfect. Clearer than any text example I can think of....

Where should I store the actual image? Should I put it in `/repo/` with the readme image, or should it go next to the migration guides? Possibly in its own folder?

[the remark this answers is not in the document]
Trouv: Yeah put it in a new directory next to the migration guides. Check out how it's done for the "tile based game" tutorial: https://github.com/Trouv/bevy_ecs_ldtk/blob/main/book/src/tutorials/tile-based-game/create-your-ldtk-project.md

there's an `images` directory there and the markdown is pointing to it w/ a title and relative path like:
```
![world-layout](images/world-layout.png)
```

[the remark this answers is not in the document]
Koopa1018: Sweet, thanks. Should've known there was a better example to check!

Koopa1018:
Alright, I've done what I can (including updating the migration guide to the new version). The rest is in your hands.... See you in 0.10 🖖

Koopa1018:
Arright, so I reorganized the platformer example so I could make heads or tails of it. Now there's a concrete problem in front of me: I'm not sure how to spawn colliders as children within the `LdtkEntity` framework. If I had an `Entity` ref, no problem, but...

Think I'm gonna have to leave this as draft until #177 finally hits. Until then, might as well submit the reorganized platformer example as a separate PR so it doesn't become hell to merge down the road....

Trouv:
Thanks for all your hard work on this. I have a lot more time for this project lately and I'm still interested in getting this PR in. Sorry for the wait.

It could use a rebase, your platformer refactor _has_ gotten in in the meantime, and some of the migration guide stuff will need to refer to newer versions of the crate (0.13 and 0.14 instead of 0.9 and 0.10).

> Now there's a concrete problem in front of me: I'm not sure how to spawn colliders as children within the LdtkEntity framework. If I had an Entity ref, no problem, but...

I'm not 100% sure what this has to do with this PR, is it a blocker? normally for things like this you do have to do post-processing instead of just `LdtkEntity` + macros

Koopa1018:
Hey, sorry I took so long to get back to this!

I've been out of the Bevy ecosystem for a while. The new Sprite architecture (Anchor being a requirement of Sprite, rather than a field) threw me for a loop, since the anchor can no longer be assigned in Sprite's `bundle_entity` function...

I'm inclined to think the simplest way to make this work would be with a post-process function (that overwrites any freshly `Added<Anchor>`), but that also has the drawback of overwriting user-defined Anchors. I could design it to only overwrite non-`default()` Anchors, but that'd still overwrite any Anchor which is equal to `default()`. That'd be a weird, counterintuitive edge case; definitely not ideal to put that burden on users.
(EDIT: That'd also be undesirable for traitless spawning, wouldn't it....)

Branch currently has the post-process system, but I'd be happy to look into the derive macro and see what I can do to make anchors even more automatic.

(I also want to write a test case to verify that this postpro system actually works, but integration testing is something I'm not super comfortable with yet.)

---

W.r.t. spawning colliders as children. I believe the problem was that Rapier colliders couldn't be offset to align with pivot-adjusted sprites except by spawning them as children. I'll try post-processing systems.

Koopa1018:
Dug into the macro. Looks like the best I can do is add an `#[anchor]` field attribute like the `#[sprite]` one. Would've liked to add a warning if a sprite has no anchor, but alas, manual compiler warnings are a nightly-only feature....

I suppose that might be for the best. If users have to manually tag their anchor fields, this whole pivot feature is effectively opt-in, which makes backwards compatibility lots easier.

Thinking of updating the macro and removing the post-process function. Thoughts?

Trouv:
That's an interesting problem. There are some design decisions in this plugin that really need some rethinking since the introduction of required components.

I did think about your idea. If we were to do something like that I'd suggest a smaller change, adding some
```rust
pub fn ldtk_entity_anchor(entity_instance: &EntityInstance) -> Anchor {
    todo!()
}
```
to be used like
```rust
#[derive(Bundle, LdtkEntity)]
struct MyBundle {
    #[sprite_sheet]
    sprite: Spite,
    #[with(ldtk_entity_anchor)]
    anchor: Anchor,
}
```

However, I realized technically `Transform`s have a similar problem. You don't include them on the bundle but the system that spawns your entity adds it anyway, and assumes that you would want the transform to just be its position in the LDtk world (with some nasty adjustments). It could be made to make a similar assumption and add an `Anchor`. I suppose ideally it would only add the anchor if the entity already had a sprite component, but that might not be possible to tell in the position it's in. It would overwrite user-defined anchors, just like user-defined transforms, but that's not immediately a dealbreaker in my mind. Lmk your thoughts on this, I'll think about it some more too.

Koopa1018:
Hmm, I hadn't considered using the existing annotations 😆

Looks like if we insert the anchor during `spawn_level` but *before* the call to `evaluate(...)`, we leave users the ability to overwrite anchors by defining them in `bundle_entity`. Spriteless entities will still get an Anchor, which is ugly, but probably harmless? This is the method I'd prefer--it's automatic until the user needs to think about it, then steps out of their way when they need to override it.

Wondering if it might be worth it to make a custom `EntityCommand` to remove unneeded Anchor components...a post-process system would work, but it's difficult to guarantee that our post-process system would run Entirely Before or Entirely After user post-process systems.

If we don't want to do this, then it'd probably be easiest for users if we just `impl LdtkEntity for Anchor` and have them flag it `#[ldtk_entity]`:
```
#[derive(Bundle, LdtkEntity)]
struct MyBundle {
    #[sprite_sheet]
    sprite: Spite,
    #[ldtk_entity]
    anchor: Anchor,
}
```
...rather than making them remember which exact function to plug into `#[with(...)]`.

Actually, I may end up doing `impl LdtkEntity for Anchor` anyway, just in case a user wants to be really explicit for some reason.

Trouv:
> Looks like if we insert the anchor during spawn_level but before the call to evaluate(...), we leave users the ability to overwrite anchors by defining them in bundle_entity.

Great point. I'm okay with this, and actually, it may make sense in the future to do that for `Transform`s as well. After all, the main reason I did transforms post-evaluate instead of pre-evaluate in the first place was because the old `SpriteSheetBundle` had a `Transform` on it. But now that it's a mere required component instead of part of the bundle added in `evaluate`, it too could be added pre-evaluate, and the plugin becomes slightly more flexible.

There may be some places in documentation where the behavior of `Transform`s is explained, may be worth explaining how `Anchor` also has similar behavior (but pre-evaluate). No need to make the change for `Transform` in this PR though, I'd prefer that as a separate small PR.

PR #304, opened 2024-02-23
Allow custom deserialization of ldtk files

StratusFearMe21:
This PR makes it possible to load Ldtk files that you have preprocessed, for example by re-serializing the file using `bincode` and compressing it.


StratusFearMe21:
Ooooo. I like that idea. Instead of an enum tho, could you pass a function pointer to the `LdtkProjectLoaderSettings`? Something like `fn(Vec<u8>) -> LdtkJson`?

PR #306, opened 2024-03-02
feat: add #[default] attribute to LdtkEntity and LdtkIntCell derive macros

### LINKED ISSUE #305: `#[ldtk_int_cell]`, `#[from_entity_instance]` etc. should not require a `Default` impl
Since https://github.com/Trouv/bevy_ecs_ldtk/pull/222, `Default` must be implemented to derive `LdtkEntity`, `LdtkIntCell` etc.

This does not interact well with `#[from_entity_instance]` and similar attributes, as I was using the attributes on fields where there is no reasonable way to implement `Default` (like a struct containing a `Box<dyn ...>`). With the change, I either have to:
- `impl Default` for my field to I can derive it for my `LdtkEntity` type, or
- manually `impl Default` for my `LdtkEntity` type, still requiring me to create a field instance out of thin air

While I *could* make up some value just to be able to implement `Default`, doing so would feel like a misuse of the `Default` trait.

neocturne:
Add an attribute that initializes fields using individual field Default implementations, allowing to use the derive macros again without requiring the whole struct to implement Default (when some fields should be default-initialized and others use initialization attributes like #[from_entity_instance]).

Fixes #305

### DIFF
--- a/src/app/ldtk_entity.rs
+++ b/src/app/ldtk_entity.rs
@@ -281,6 +281,37 @@ use std::{collections::HashMap, marker::PhantomData};
 ///     }
 /// }
 /// ```
+///
+/// ### `#[default]`
+///
+/// Indicates that this component or bundle should be initialized using
+/// [Default::default].
+/// This can be useful when implementing `Default` for the whole `LdtkEntity` is
+/// not easily possible, because some of the fields do not implement `Default`.
+///
+/// ```
+/// # use bevy::prelude::*;
+/// # use bevy_ecs_ldtk::prelude::*;
+/// # #[derive(Component, Default)]
+/// # struct Damage;
+/// # #[derive(Component, Default)]
+/// # struct BleedDamage;
+/// #[derive(Bundle, LdtkEntity)]
+/// pub struct Weapon {
+///     #[default]
+///     damage: Damage,
+///     #[sprite_bundle]
+///     sprite: SpriteBundle,
+/// }
+///
+/// #[derive(Bundle, LdtkEntity)]
+/// pub struct Dagger {
+///     #[ldtk_entity]
+///     weapon_bundle: Weapon,
+///     #[default]
+///     bleed_damage: BleedDamage,
+/// }
+/// ```
 pub trait LdtkEntity {
     /// The constructor used by the plugin when spawning entities from an LDtk file.
     /// Has access to resources/assets most commonly used for spawning 2d objects.
--- a/src/app/ldtk_entity.rs
+++ b/src/app/ldtk_entity.rs
@@ -285,31 +285,28 @@ use std::{collections::HashMap, marker::PhantomData};
 /// ### `#[default]`
 ///
 /// Indicates that this component or bundle should be initialized using
-/// [Default::default].
+/// [`Default::default`].
 /// This can be useful when implementing `Default` for the whole `LdtkEntity` is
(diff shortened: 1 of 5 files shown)


src/app/ldtk_entity.rs
```diff
@@ -281,6 +281,37 @@ use std::{collections::HashMap, marker::PhantomData};
...
+///     sprite: SpriteBundle,
+/// }
+///
+/// #[derive(Bundle, LdtkEntity)]
+/// pub struct Dagger {
+///     #[ldtk_entity]
+///     weapon_bundle: Weapon,
+///     #[default]
+///     bleed_damage: BleedDamage,
+/// }
```
Trouv, line 313: I think for both of these docs a more concise example demonstrating the problem this addresses should be provided. maybe something like..
```suggestion
/// # use bevy::prelude::*;
/// # use bevy_ecs_ldtk::prelude::*;
/// # mod other_crate {
/// #     #[derive(bevy::prelude::Component)]
/// #     pub struct ForeignComponentWithNoDefault;
/// # }
/// # fn custom_constructor(_: &EntityInstance) -> ForeignComponentWithNoDefault { todo!(); }
/// # #[derive(Component, Default)]
/// # struct Damage;
/// use other_crate::ForeignComponentWithNoDefault;
///
/// #[derive(Bundle, LdtkEntity)]
/// pub struct MyBundle {
///     #[default]
///     damage: Damage,
///     #[with(custom_constructor)]
///     foreign: ForeignComponentWithNoDefault,
/// }
```

src/app/ldtk_entity.rs
```diff
@@ -281,6 +281,37 @@ use std::{collections::HashMap, marker::PhantomData};
 ///     }
 /// }
 /// ```
+///
+/// ### `#[default]`
+///
+/// Indicates that this component or bundle should be initialized using
+/// [Default::default].
```
Trouv, line 288: NIT: kinda following this convention for intradoc links
```suggestion
/// [`Default::default`].
```

src/app/ldtk_int_cell.rs
```diff
@@ -154,6 +154,35 @@ use std::{collections::HashMap, marker::PhantomData};
 ///     damage: Damage,
 /// }
 /// ```
+///
+/// ### `#[default]`
+///
+/// Indicates that this component or bundle should be initialized using
+/// [Default::default].
```
Trouv, line 161: same NIT:
```suggestion
/// [`Default::default`].
```

Trouv (commented):
I'm suddenly remembering why I originally made the plugin use the field's `Default` instead of the bundle's. I think this is a great compromise, thanks for your work/design on it. I only have a few docs-related suggestions

neocturne:
Is this PR still missing anything?

neocturne:
Friendly ping, I'd love to see my change in the next release!

Trouv:
Sorry for being unresponsive on this. I'd like to see it in the next release too.

I think the only thing that was missing in my mind was a similar change to the example code in `ldtk_int_cell.rs` that I suggested for `ldtk_entity.rs` [here](https://github.com/Trouv/bevy_ecs_ldtk/pull/306#discussion_r1550851434). I'll make this change myself tomorrow unless you beat me to it. I appreciate the patience on this and will take responsibility to implement any remaining changes to make up for being so absent

PR #318, opened 2024-04-19
Remove unused asset server param

jgayfer:
This param wasn't being "used" in this example code.


Trouv (approved):
Nice catch, thanks for the fix!

Issue #94, opened 2022-06-12
Tilemap shows wrong tiles when using spacing and atlas


svenallers:
Hi,
I currently have the issue that the rendered tilemap does not look at all like the one I created.

Here is how it looks in LDtk:
<img width="1792" alt="ldtk" src="https://user-images.githubusercontent.com/7449886/173254281-12a0cbf7-d793-4916-922b-7af67941ae31.png">
 
And here is how it is rendered when starting my game:
<img width="1283" alt="bevy" src="https://user-images.githubusercontent.com/7449886/173254311-306bf52d-6abf-4626-8140-c4e43a41a996.png">

The tileset is using a spacing of 2px and therefore I enabled the atlas feature. My code currently looks like this:

_main.rs_
```rust
use bevy::prelude::*;
use bevy_ecs_ldtk::prelude::*;

fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_plugin(LdtkPlugin)
        .add_startup_system(setup)
        .insert_resource(LevelSelection::Index(0))
        .run();
}

fn setup(mut commands: Commands, asset_server: Res<AssetServer>) {
    commands.spawn_bundle(OrthographicCameraBundle::new_2d());

    commands.spawn_bundle(LdtkWorldBundle {
        ldtk_handle: asset_server.load("racing-track.ldtk"),
        ..Default::default()
    });
}
```

_cargo.toml_
```toml
[package]
edition = "2021"
name = "bevy-racer"
version = "0.1.0"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
bevy = "0.7"
bevy_ecs_ldtk = {version = "0.3", features = ["atlas"]}
```

A link to the full codebase, having the issue including the ldtk file and resources:  https://github.com/svenallers/bevy-racer/commit/1e7e76d63d7247ea3a73b31679bfa75717afdd16

Could also be that the issue is on me, but I have no clue how to solve it.

svenallers:
I just dug into the calculation of the position of the tiles in the tilesets and it seems that LDtk is handling indices differently, than`bevy_ecs_tilemap` when the size of the sprite sheet does not align perfectly with the tile size (meaning it is not n * tile_size but sth. different).

Seems that LDtk adds an index for the "not complete" tile. See:
[...]

PR #351, opened 2025-03-03
docs: document spacing and padding effect on the rendered tilemap #94

azthec:
Adding the behaviour of `bevy_ecs_tilemap` and spacing / padding to the book.

Feel free to change the wording as you will

### DIFF
--- a/book/src/tutorials/tile-based-game/spawn-your-ldtk-project-in-bevy.md
+++ b/book/src/tutorials/tile-based-game/spawn-your-ldtk-project-in-bevy.md
@@ -52,7 +52,9 @@ fn main() {
 }
 ```
 
-Now, run the game with `$ cargo run --release` to see your first level spawning in Bevy!
+Now, run the game with `$ cargo run --release` to see your first level spawning in Bevy! Is the rendered level different than shown in LDtk? [^1])
+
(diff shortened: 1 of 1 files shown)


book/src/tutorials/tile-based-game/spawn-your-ldtk-project-in-bevy.md
```diff
@@ -52,7 +52,9 @@ fn main() {
 }
 ```
 
-Now, run the game with `$ cargo run --release` to see your first level spawning in Bevy!
+Now, run the game with `$ cargo run --release` to see your first level spawning in Bevy! Is the rendered level different than shown in LDtk? [^1])
+
+[^1]: Due to a difference in the handling of spacing and padding between `bevy_ecs_tilemap` and LDtk spacing is not perfectly supported. This can be resolved by having the value of padding and spacing be equal, for example both 0, or both 1 and so on. Previous versions of `bevy_ecs_tilemap` require the `atlas` feature flag enabled for WASM support and also for tile spacing to work with Tile and AutoTile layers.
```
Trouv, line 57: Thanks for your contribution! Do you think you could instead make a new explanation chapter called "Limitations" with a section for this? Then you could link that page here instead of using a footnote.

It'll be a small chapter for now,  but I would like to have something I can easily link to when users have this issue in the future. It would also serve as a place to put other explanations of limitations and workarounds as users encounter them.

azthec: Can you take a second look? I think this is what you meant

Trouv (approved):
LGTM, thanks again

PR #353, opened 2025-04-12
fix: Migrate to 0.16.0-rc.2

ethanuppal:
I followed the [0.15 -> 0.16 migration guide](https://bevyengine.org/learn/migration-guides/0-15-to-0-16/) and mechanically applied the changes.

There was one part where `rustc` complained about type mismatch on a `match`. I assume this is a new error/lint where there used to be some implicit to-unit casting. I can remove this if needed, but I think it improves clarity/correctness (and seems to be required to be compiled by my (latest stable) toolchain).

I also used `unwrap_or_default()` on an `Option<Vec<_>>`. I can replace this with an `if let` if you are worried about `clone` performance in that spot.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,9 +15,9 @@ members = ["macros"]
 
 [dependencies]
 bevy_ecs_ldtk_macros = { version = "0.11.0", optional = true, path = "macros" }
-bevy_ecs_tilemap = { version = "0.15.0", default-features = false }
-bevy = { version = "0.15", default-features = false, features = [
-    "bevy_sprite",
+bevy_ecs_tilemap = { version = "0.16.0-rc.1", default-features = false }
+bevy = { version = "0.16.0-rc.2", default-features = false, features = [
+  "bevy_sprite",
 ] }
 derive-getters = "0.3.0"
 serde = { version = "1.0", features = ["derive"] }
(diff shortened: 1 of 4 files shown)


andriyDev:
Hi, thanks for doing this. Could you update this PR to 0.16.0? I'd like to patch this into my project, but I would like to make sure it doesn't include rc.2

ethanuppal:
I'll probably do it when I update my project to 0.16

nfagerlund:
Hello! I took this branch as a starting point and updated to bevy 0.16.0, then fixed most of the deprecation warnings, then upgraded dependencies and fixed errors in examples/ and got all the example apps running right again (I think). 

The result is here: 

https://github.com/nfagerlund/bevy_ecs_ldtk/tree/may25-bevy016

Definitely not mergeable at the moment, because it's got the bevy_ecs_tilemap dep (and the bevy_rapier dev-dep for the platformer example) pinned to git SHAs, as neither of them seem to have final 0.16 releases out yet and there are important fixes on main that aren't in their latest RC tags. But it might be useful to others who just wanna update their personal projects. 

The deprecation warnings I didn't fix were all `Handle::weak_from_u128` invocations in tests; I wasn't sure _exactly_ what guarantees those tests wanted to uphold with their fake UUIDs (and was eager to get back to fixing my actual game), so I left it alone.

Fincap:
FYI, `bevy_ecs_tilemap` has been updated to Bevy 0.16: https://github.com/StarArawn/bevy_ecs_tilemap/pull/613

PR #365, opened 2025-09-21
Use let-else for systems to reduce nesting.

andriyDev:
Previously, these systems were very heavily nested. Thanks to let-else! Now there's **much** less nesting, making it easier to grasp the control flow!


Trouv:
when in doubt, clear cache

PR #366, opened 2025-09-21
Reduce the number of commands issued when inserting/spawning.

andriyDev:
Previously multiple insert calls would result in multiple archetype moves, especially when spawning. Bundling up these insert commands makes them cheaper!

This also leaves the entities in a more consistent state for observers.


src/level.rs
```diff
@@ -474,7 +478,8 @@ pub fn spawn_level(
                 {
                     let layer_entity = commands.spawn_empty().id();
 
-                    let tilemap_bundle = if layer_instance.layer_instance_type == Type::IntGrid {
+                    let mut tilemap_bundle = if layer_instance.layer_instance_type == Type::IntGrid
```
Trouv: The only suggestion I have here is to avoid the mutability of this variable. I _think_ everything needed to calculate the transform of the layer can just be done above this line, that way the tilemap bundle can just be initiated with that transform instead of updated to it.

PR #367, opened 2025-09-21
Only generate the worldly_set once per frame and only when spawning.

andriyDev:
Previously if multiple levels were being loaded at once, we would need to rebuild this set of worldly entities multiple times. In addition, if a level was missing its level data, we would still build the worldly_set for it! This fixes both issues.

Note: This PR is based off #365, since these PRs will conflict.


Trouv (approved):
seems to work well. Usually I'd prefer avoiding the mutable variable, but doing so here would be somewhat of a lift

PR #368, opened 2025-09-21
Use the Respawn component to determine whether scenes should be spawned instead of guessing using the children.

andriyDev:
Previously we would decide whether to process a level based on whether it has any children. In some situations, however a processed level can still end up with no children. In addition, this makes it harder to wait for some assets to be loaded (e.g., waiting for external levels to be loaded).

This is based on #365 and #366.


Trouv:
all except the last commit here has been merged right?

andriyDev:
Yup give me a few minutes to rebase this!

PR #404, opened 2026-06-25
Bevy 0.19

neocturne:
- [x] Waiting for https://github.com/StarArawn/bevy_ecs_tilemap/pull/634
- [x] Depends on https://github.com/Trouv/bevy_ecs_ldtk/pull/405, as there is not even a git version of bevy_rapier with support for 0.19, and there may not be for some time due to the dependency incompatibilities noted in https://github.com/dimforge/bevy_rapier/pull/694
- [x] Decide how to deal with inconsistent sprite ordering (caused by the avian migration, not Bevy 0.19)

~With the platformer example I'm seeing inconsistent sprite ordering between the player and the chests after the update to 0.19. See https://github.com/Trouv/bevy_ecs_ldtk/pull/404#issuecomment-4805200384.~

I've also given this a superficial test with a simple personal project. I didn't notice any obvious issues here.

### DIFF
--- a/src/systems.rs
+++ b/src/systems.rs
@@ -340,7 +340,7 @@ pub fn clean_respawn_entities(world: &mut World) {
             other_ldtk_levels,
             worldly_entities,
             mut level_events,
-        ) = system_state.get_mut(world);
+        ) = system_state.get_mut(world).unwrap();
 
         for world_children in ldtk_worlds_to_clean.iter() {
             for child in world_children
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,8 +15,8 @@ members = ["macros"]
 
 [dependencies]
 bevy_ecs_ldtk_macros = { version = "0.14.0", optional = true, path = "macros" }
-bevy_ecs_tilemap = { version = "0.18.0", default-features = false }
-bevy = { version = "0.18", default-features = false, features = [
+bevy_ecs_tilemap = { version = "0.19.0", default-features = false }
+bevy = { version = "0.19", default-features = false, features = [
     "bevy_sprite",
 ] }
 derive-getters = "0.5.0"
@@ -28,11 +28,11 @@ paste = "1.0"
 derive_more = "0.99.20"
 
 [dev-dependencies]
-bevy = "0.18"
-avian2d = "0.6"
+bevy = "0.19"
+avian2d = "0.7"
 fake = { version = "4.4", features = ["uuid"] }
 rand = "0.9"
-bevy-inspector-egui = "0.36.0"
+bevy-inspector-egui = "0.37.0"
 
 [features]
 default = ["derive", "render", "internal_levels"]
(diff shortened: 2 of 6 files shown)


neocturne:
> With the platformer example I'm seeing inconsistent sprite ordering between the player and the chests after the update to 0.19. I have not investigated the cause yet.

Hmm, I guess the draw order is basically undefined for sprites with the same z coordinate? Still a regression from 0.18, where the order was stable, rather than changing when moving. I also can't find anything related in the release notes or migration guide.

I see the following options:

- Fix in the library: apply a unique z offset to all entities inside a layer
- Fix in the example: set a z offset for the player
- Accept the inconsistent behavior
- Report the change in behavior to bevy?

src/systems.rs
```diff
@@ -340,7 +340,7 @@ pub fn clean_respawn_entities(world: &mut World) {
             other_ldtk_levels,
             worldly_entities,
             mut level_events,
-        ) = system_state.get_mut(world);
+        ) = system_state.get_mut(world).unwrap();
```
Trouv, line 343: I suppose panicking is valid here, since this would only really fail if the `SystemState` generic is breaking some bevy rules, and we know it isn't. But, could we use an `expect` instead?

Trouv (commented):
Thanks for this, so far I've only reviewed `src`. Curious how difficult it would be to split this up into a PR for the avian migration and the bevy 0.19 migration? I don't mind merging this with a git dependency on rapier

src/systems.rs
neocturne: I thought about using `expect`, but the example in the bevy repo uses unwrap too for `SystemState::get_mut`, and the panic will include the [SystemParamValidationError](https://docs.rs/bevy/latest/bevy/ecs/system/struct.SystemParamValidationError.html).

neocturne:
Avian migration moved to https://github.com/Trouv/bevy_ecs_ldtk/pull/405 and PR description updated.

Trouv:
I see the issue you're talking about w/ the sprite ordering, though I think it had something to do with the avian migration? Not bevy 0.19, I'm seeing it main now too.

Regardless I think it's acceptable for now. It kinda makes sense to me that ordering between equal z-value entities would be undefined. Idk if bevy_ecs_ldtk should have an opinion on it, I think if users want to explicitly make the player on top, then add .1 to its z in post-processing. I'd be happy accepting a system for that in the platformer example in another PR

neocturne:
Oh, you are right! I thought I hadn't seen in with avian on 0.18, but I guess I didn't properly pay attention. Kinda weird that the physics engine has an effect on the sprite ordering when it should handle only two dimensions, I'll try to figure out what is going on.

neocturne:
Hmm, using bevy-inspector-egui I see additional components being added to entities while they are colliding. Maybe archetype changes can result in a different draw order for equal z coordinates?

mnmaita:
@neocturne `bevy_ecs_tilemap` got updated to 0.19 FYI 😄

neocturne:
Great, PR is updated now.

Trouv (approved):

PR #405, opened 2026-06-26
Port platformer example from bevy_rapier to avian

neocturne:
Being a native bevy plugin, avian provides support for new bevy versions quickly and is not plagued by dependency incompatibilities (like requiring rapier and bevy to use the same version of glam).

Prepares for the Bevy 0.19 update in https://github.com/Trouv/bevy_ecs_ldtk/pull/404. As noted in https://github.com/dimforge/bevy_rapier/pull/694, rapier skipped the glam version used by Bevy 0.19, so making them work together is cumbersome (and not even using a git version of bevy_rapier would work). Using avian instead of bevy_rapier allows us to update bevy without waiting for a solution and avoids similar problems in the future.

As far as I can tell, nothing about the game feels significantly different with this change applied. One ugly hack was necessary to make it work correctly, marked by a comment.

### DIFF
--- a/examples/platformer/main.rs
+++ b/examples/platformer/main.rs
@@ -4,7 +4,7 @@
 use bevy::prelude::*;
 use bevy_ecs_ldtk::prelude::*;
 
-use bevy_rapier2d::prelude::*;
+use avian2d::prelude::*;
 
 mod camera;
 mod climbing;
@@ -22,10 +22,8 @@ mod walls;
 fn main() {
     App::new()
         .add_plugins(DefaultPlugins.set(ImagePlugin::default_nearest()))
-        .add_plugins((
-            LdtkPlugin,
-            RapierPhysicsPlugin::<NoUserData>::pixels_per_meter(100.0),
-        ))
+        .add_plugins((LdtkPlugin, PhysicsPlugins::default()))
+        .insert_resource(Gravity(Vec2::new(0.0, -2000.0)))
         .insert_resource(LevelSelection::Uid(0))
         .insert_resource(LdtkSettings {
             level_spawn_behavior: LevelSpawnBehavior::UseWorldTranslation {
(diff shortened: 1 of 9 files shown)


examples/platformer/walls.rs
```diff
@@ -179,6 +177,13 @@ pub fn spawn_wall_collision(
                 });
             }
         });
+
+        // Hack: The wall colliders spawn one frame later than the player.
+        // Only start the physics simulation after that, so the player doesn't
+        // end up in the floor.
+        commands.queue(|world: &mut World| {
+            world.resource_mut::<Time<Physics>>().unpause();
+        });
```
Trouv: Does it work to do this in a new system on `LevelEvent::Transformed`? I'd like to keep this system concerned about spawning wall colliders if possible.

examples/platformer/main.rs
```diff
@@ -22,10 +22,13 @@ mod walls;
 fn main() {
     App::new()
         .add_plugins(DefaultPlugins.set(ImagePlugin::default_nearest()))
-        .add_plugins((
-            LdtkPlugin,
-            RapierPhysicsPlugin::<NoUserData>::pixels_per_meter(100.0),
-        ))
+        .add_plugins((LdtkPlugin, PhysicsPlugins::default()))
+        .insert_resource(Gravity(Vec2::new(0.0, -2000.0)))
+        .insert_resource({
```
Trouv: Might be good to leave a comment about the temporary pause, and mention the system it's unpaused in.

Trouv (commented):
Thanks for the split, `main`'s git log will look better when squashing these individually.

examples/platformer/walls.rs
neocturne: Moved to a new system on `LevelEvent::Transformed`.

examples/platformer/main.rs
neocturne: Moved the physics pause to `setup` and also moved the comment about the pause.

Trouv (approved):

PR #408, opened 2026-09-23
Add support for BSN to LdtkProjectHandle

jgayfer:
BSN ([Bevy Scene Notation](https://docs.rs/bevy/latest/bevy/scene/index.html)) is the new kid on the block in Bevy.

However in order to use it, our components need to be setup to use required components, as well as ensure we're implementing `FromTemplate` for any components that require world/spawn context (like asset handles).

I've updated the basic example to showcase how this cleans up setup. We could certainly update more, but want to keep it tightly scoped for now :)

---

_No AI was used in the making of this PR._


src/components/mod.rs
```diff
@@ -330,7 +330,15 @@ pub(crate) struct EntityInstanceBundle {
...
 /// [*Anatomy of the World*](https://trouv.github.io/bevy_ecs_ldtk/v0.15.0/explanation/anatomy-of-the-world.html) <!-- x-release-please-version -->
 /// chapter of the `bevy_ecs_ldtk` book.
-#[derive(Debug, Default, Clone, Component, Reflect, Deref, DerefMut)]
+#[derive(Debug, Default, Clone, Component, Reflect, Deref, DerefMut, FromTemplate)]
+#[require(
+    LevelSet,
+    Transform,
+    GlobalTransform,
+    Visibility,
+    InheritedVisibility,
```
Trouv: I think it's a little cleaner to not require `GlobalTransform`/`InheritedVisibility`/`ViewVisibility` since they are already required by `Transform`/`Visibility`

examples/basic.rs
```diff
@@ -7,19 +7,17 @@ fn main() {
...
-    commands.spawn(LdtkWorldBundle {
-        ldtk_handle: asset_server.load("my_project.ldtk").into(),
-        ..Default::default()
-    });
+fn scene() -> impl Scene {
+    bsn! {
+        Camera2d
+        LdtkProjectHandle { handle: "my_project.ldtk" }
+    }
 }
```
Trouv: Could you equivalently update this in the `traitless` example?

Edit: this repo doesn't have a code example in the README.md

src/components/mod.rs
jgayfer: TIL! That's nice to know.

